PR #3, opened 2017-04-08
[New] add `isValidElement` entry point

### LINKED ISSUE #1: export `isValidElement`?
Could there be an entry point for `isValidElement`? I'd like to use it without depending on React :-)

ljharb:
Fixes #1.

I think a much better approach, however, would be to make `isValidElement` a separate package too, like `prop-types` - happy to close this PR in place of that :-)

### DIFF
--- a/__tests__/isValidElement-test.js
+++ b/__tests__/isValidElement-test.js
@@ -1,0 +1,31 @@
+'use strict';
+
+var React = require('react');
+var isValidElement = require('../isValidElement');
+
+describe('isValidElement', () => {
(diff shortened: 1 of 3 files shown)


aweary:
Per discussion in https://github.com/facebook/prop-types/issues/1 it seems we all agree that while this might be useful, this wouldn't be the right place for it

ljharb:
@aweary where would i discuss there being a new package for it?

aweary:
@ljharb the best place would probably be an issue in the React repo. You'll get the most eyes on it there.

PR #13, opened 2017-04-10
Remove build files from git tracking

aweary:
There's no good reason that I can see to track `prop-types.js` and `prop-types.min.js` in git.


jarstingstall:
Having these builds is helpful for projects that want to include the UMD builds and avoid the CDN request.

Daniel15:
Thank you! I was going to submit a diff for this but I'm glad someone else got to it first 😛

PR #15, opened 2017-04-10
improve union type checker error message

johann-sonntagbauer:
PropTypes type checking: Add additional check to UnionType checker to ensure only functions are used as argument of oneOfType.

This is especially handy to catch typos during PropType definition like for example:
```javascript
Greeting.propTypes = {
  name: React.PropTypes.oneOfType([React.PropTypes.sting, React.PropTypes.func])
};
```

The pull request will print a warning 
`Invalid argument supplied to oneOfType, expected an array containing check functions.`

Pull Request taken over from React main Repository
https://github.com/facebook/react/pull/8823

### DIFF
--- a/__tests__/PropTypesDevelopmentReact15.js
+++ b/__tests__/PropTypesDevelopmentReact15.js
@@ -859,6 +859,24 @@ describe('PropTypesDevelopmentReact15', () => {
       typeCheckPass(PropTypes.oneOf(PropTypes.string, PropTypes.number), []);
     });
 
+    it('should warn when setup with invalid checker types', () => {
+      spyOn(console, 'error');
+
+      PropTypes.oneOfType([undefined])
+
+      expect(console.error).toHaveBeenCalled();
+      expect(console.error.calls.argsFor(0)[0])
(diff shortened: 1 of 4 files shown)


factory.js
```diff
@@ -326,6 +326,13 @@ module.exports = function (isValidElement) {
       return emptyFunction.thatReturnsNull;
     }
 
+    for (var i = 0; i < arrayOfTypeCheckers.length; i++) {
+      if (typeof arrayOfTypeCheckers[i] !== 'function') {
+        warning(false, 'Invalid argument supplied to oneOfType, expected an array containing check functions.');
```
aweary: What do you think about including the index and type in the warning?

> Invalid argument supplied to oneOfType, expected an array containing check functions but received null at index 2

johann-sonntagbauer: good idea, will add the information

johann-sonntagbauer:
@aweary introduced the change request of yours. Message is now much clearer.

johann-sonntagbauer:
Tried to rebase onto current master. Hope I interpreted the changes in the right way. Please ping me when the master is stable enough to rebase again.

johann-sonntagbauer:
@aweary thx for review. Changes are good.

PR #20, opened 2017-04-11
Revert accidental breaking change for 15.5

gaearon:
There's a bunch of changes to unpack here but basically:

* Created a separate text fixtures for development, production in 15, and production with dependency on standalone `prop-types` package.
* Their intended behavior is this:
  - `PropTypesDevelopmentReact15`: When you depend on `React@15.PropTypes` in development, it runs type checks and shows warnings.
  - `PropTypesProductionReact15`: When you depend on `React@15.PropTypes` in production, everything is a no-op. **It doesn't run the type checks but the code is still there. Calling validators doesn’t throw. This fixes https://github.com/reactjs/prop-types/issues/17.**
  - `PropTypesDevelopmentStandalone` and `PropTypesProductionStandalone:`: When you depend on `prop-types`, it runs type checks and shows warnings in development. Manually calling validators throws. (Just like we warned in earlier releases of React itself.) This is not a breaking change because migrating to `prop-types` is an intentional decision, and requires you to update your code. We could call it out more explicitly but it’s explained in README.

I also cherry-picked @aweary’s two commits for reducing the browserify bundle from https://github.com/reactjs/prop-types/pull/18 so that I could verify my changes. Production UMD bundle shows dead code is eliminated correctly.

When we release 16, we can drop the `React15` tests, delete the `factory` indirection and bump the major version.


flarnie:
Great commit message - reading the code now. And let me know if there are any other checks I should do to verify this - pulling down the change and running fixtures, or anything like that.

gaearon:
I don't really have any ideas for testing this except creating a project, dropping this version of `prop-types` there, and verifying proptypes work as expected in prod builds (according to the above matrix).

PR #25, opened 2017-04-12
Add a hack for React 0.14 compatibility

gaearon:
I am looking at the stack to see if we're plausibly calling from React 0.14, and opting into the old behavior in this case.

I'm sorry for doing this.

Fixes https://github.com/reactjs/react-redux/issues/669


gaearon:
Meh. This is too fragile.

PR #29, opened 2017-04-14
Added a bit more info to the warning when failing a type union check

thaggie:
I found the warning from type union failures to be uninformative and so have tried to improve them. 
I've changed the message to append the failure messages from each of the types on new lines so rather than just seeing that the prop type was wrong you can see what it was trying to match against.

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -321,14 +321,17 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
     }
 
     function validate(props, propName, componentName, location, propFullName) {
+      var messages = [];
       for (var i = 0; i < arrayOfTypeCheckers.length; i++) {
         var checker = arrayOfTypeCheckers[i];
-        if (checker(props, propName, componentName, location, propFullName, ReactPropTypesSecret) == null) {
+        var error = checker(props, propName, componentName, location, propFullName, ReactPropTypesSecret);
+        if (error == null) {
           return null;
         }
+        messages.push(error.message);
       }
 
-      return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` supplied to ' + ('`' + componentName + '`.'));
+      return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` supplied to ' + ('`' + componentName + '`, failed matches:\n' + messages.join('\n')));
     }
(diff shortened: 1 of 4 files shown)


zpao:
I don't know if the CLA bot is going to come around (it might after I comment…). If we continue with this we'll the CLA signed (https://code.facebook.com/cla). If you run into any issues, please email cla@fb.com

thaggie:
I'm waiting on my company approving the CLA - they're not used to this sort of thing so it's taking them a while.

thaggie:
@zpao turns out I'd already signed it.

factoryWithTypeCheckers.js
```diff
@@ -321,14 +321,17 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
-        if (checker(props, propName, componentName, location, propFullName, ReactPropTypesSecret) == null) {
+        var error = checker(props, propName, componentName, location, propFullName, ReactPropTypesSecret);
+        if (error == null) {
           return null;
         }
+        messages.push(error.message);
       }
 
-      return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` supplied to ' + ('`' + componentName + '`.'));
+      return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` supplied to ' + ('`' + componentName + '`, failed matches:\n' + messages.join('\n')));
```
ljharb: does this mean that the text "failed matches" will now show up for every propType error, even if there's only one error message?

maybe it'd be better to conditionally add the new text only when `messages.length > 1`?

thaggie:
@aweary I did - see my response comment.

aweary:
@thaggie sorry, I don't see any response or commits related to what @ljharb mentioned? Specifically the feedback left in https://github.com/facebook/prop-types/pull/29#discussion_r121272637

thaggie:
@aweary I think there's maybe my comment wasn't "published" I now notice there's a "Pending" next to my name.

This is what it looks like on for me:
screen shot 2017-07-10 at 7 38 07 pm

I didn't make a change because it's not really a valid use case. Technically it is possible to have a single type in a type union but I'd have thought it better to keep the code logic free rather than adding in a bunch of logic to cope with it.

I'm certainly willing to change it to add that logic - but if I was responsible for the code base I would not want it there.

PR #35, opened 2017-04-20
Add API reference to README

flarnie:
We took the API reference from the React Docs guide on 'Typechecking
with PropTypes' and added it to this README.

Once this lands we can remove it from those docs and link to these docs.

Why? So that we make it clear that this is a separate module that can be
used with other libraries as well, and not just a feature of React. Also
hopefully this will make things more clear.

We will also open a `good-first-bug` issue for someone to improve this
README and turn it into a more complete API reference, with examples
using PropTypes with other libraries.

Fixes https://github.com/facebook/react/issues/9467

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,12 +20,97 @@ If you prefer a `<script>` tag, you can get it from `window.PropTypes` global:
 ```html
 <!-- development version -->
 <script src="https://unpkg.com/prop-types/prop-types.js"></script>
- 
+
 <!-- production version -->
 <script src="https://unpkg.com/prop-types/prop-types.min.js"></script>
 ```
 
 ## Usage
+
+Here is an example documenting the different validators provided:
+
+```javascript
+import PropTypes from 'prop-types';
+
+MyComponent.propTypes = {
+  // You can declare that a prop is a specific JS primitive. By default, these
+  // are all optional.
+  optionalArray: PropTypes.array,
+  optionalBool: PropTypes.bool,
+  optionalFunc: PropTypes.func,
+  optionalNumber: PropTypes.number,
+  optionalObject: PropTypes.object,
+  optionalString: PropTypes.string,
+  optionalSymbol: PropTypes.symbol,
+
+  // Anything that can be rendered: numbers, strings, elements or an array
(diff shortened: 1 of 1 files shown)


flarnie:
That is worthwhile, I might do it later. If I have time now I will, just don't think it should block anything else.

README.md
```diff
@@ -20,13 +20,98 @@ If you prefer a `<script>` tag, you can get it from `window.PropTypes` global:
...
 <script src="https://unpkg.com/prop-types/prop-types.min.js"></script>
 ```
 
 ## Usage
 
+Here is an example documenting the different validators provided:
+
+```javascript
+import PropTypes from 'prop-types';
+
```
gaearon: Maybe worth defining a dummy `MyComponent` so that it doesn’t look like part of the API?

README.md
```diff
@@ -20,13 +20,98 @@ If you prefer a `<script>` tag, you can get it from `window.PropTypes` global:
 ```html
 <!-- development version -->
 <script src="https://unpkg.com/prop-types/prop-types.js"></script>
- 
+
 <!-- production version -->
 <script src="https://unpkg.com/prop-types/prop-types.min.js"></script>
 ```
 
 ## Usage
```
gaearon: Maybe worth adding a sentence about the purpose of this library? Something like

>You can use `prop-types` to document the intended types of properties passed to components. React (and potentially other libraries—see the `checkPropTypes()` reference below) will check props passed to your components against those definitions, and warn in development if they don’t match. Here is an example of using `prop-types` with a React component:

README.md
flarnie: That makes sense - I'll rewrite and add to this so that it would make sense to someone who has not got the context of React.

README.md
flarnie: Sounds good - will fix.

README.md
```diff
@@ -2,6 +2,11 @@
 
 Runtime type checking for React props and similar objects.
 
+You can use prop-types to document the intended types of properties passed to
```
gaearon: It would be nice to use backticks for APIs like `checkPropTypes()`.

PR #41, opened 2017-04-27
Add exact type checker from react native

thejameskyle:
I adapted [`createStrictShapeTypeChecker`](https://github.com/facebook/react-native/blob/c3b25c9059f57ee8df5505628ff6221f11cf9234/Libraries/Utilities/createStrictShapeTypeChecker.js) from react native into a `PropTypes.exact` checker.

This operates exactly the same way as `PropTypes.shape` except it will warn on additional keys:

```js
check(PropTypes.shape({}), { foo: 1 }); // Okay.
check(PropTypes.exact({}), { foo: 1 }); // Warning!
```

This is a runtime version of [Flow's exact object types](https://flow.org/en/docs/types/objects/#toc-exact-object-types):

```js
type shape = { ... };
type exact = {| ... |};
```

### DIFF
--- a/__tests__/PropTypesDevelopmentReact15.js
+++ b/__tests__/PropTypesDevelopmentReact15.js
@@ -876,7 +876,7 @@ describe('PropTypesDevelopmentReact15', () => {
         );
         console.error.calls.reset();
       }
-      
+
       typeCheckPass(PropTypes.oneOf(PropTypes.string, PropTypes.number), []);
     });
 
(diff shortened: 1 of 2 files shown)


gaearon:
I reverted in https://github.com/facebook/prop-types/pull/80 because the review missed a few breakages related to how this repo is setup (sorry they're not obvious 😞 ).

Happy to take it back once they're addressed!

PR #43, opened 2017-04-29
Added some missing syntax highlighting.

oyeanuj:
Hope it makes it a little bit easier on the eye :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ definitions, and warn in development if they don’t match.
 
 ## Installation
 
-```
+```shell
 npm install --save prop-types
 ```
 
@@ -37,7 +37,7 @@ commonly used with React components.
 Here is an example of using PropTypes with a React component, which also
 documents the different validators provided:
 
-```javascript
+```jsx
 import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,7 +37,7 @@ commonly used with React components.
 Here is an example of using PropTypes with a React component, which also
 documents the different validators provided:
 
-```javascript
+```jsx
 import React from 'react';
```
flarnie: Looks like we don't actually use jsx in this code block, but in case we did in the future I think this change still makes sense.

README.md
```diff
@@ -236,7 +236,7 @@ React will automatically check the propTypes you set on the component, but if
 you are using PropTypes without React then you may want to manually call
 `PropTypes.checkPropTypes`, like so:
 
-```
+```js
```
flarnie: Why 'jsx' above and just 'js' here? This is still better than no syntax highlighting of course.

oyeanuj: For some reason, 'jsx' syntax highlighting works much better than 'js' syntax highlighting for objects. So I tend to opt for that, as well as incase one need to have JSX in the code snippet in future as well.

PR #51, opened 2017-05-13
Friendlier message when using a type checker that is not a function

### LINKED ISSUE #33: Why PropType.boolean is function
```javascript: 
Modal.propTypes = {visible: PropTypes.boolean}
```

> Warning: Failed prop type: Modal: prop type `visible` is invalid; it must be a function, usually from React.PropTypes.

I think that the boolean type should be called 'boolean'
You can add boolean as alias for bool, for expected behavior PropTypes

I think if undefined PropType shouldn't be function

gouegd:
Just a text change, adds a stronger hint as to the most common issue.
This tries to prevent confusion as discussed in #33 

Sample:

```
> PropTypes.checkPropTypes({pname: PropTypes.stringzz}, {pname: false}, 'prop', 'MyCompo')
Warning: Failed prop type: MyCompo: type specification of prop `pname` is invalid; the type checker must be a function but is `undefined`. This often happens because of typos such as `React.PropTypes.function` instead of `React.PropTypes.func`.
```

Fixes #33.

### DIFF
--- a/checkPropTypes.js
+++ b/checkPropTypes.js
@@ -53,7 +53,8 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
           if (typeof typeSpecs[typeSpecName] !== 'function') {
             var err = Error(
               (componentName || 'React class') + ': ' + location + ' type `' + typeSpecName + '` is invalid; ' +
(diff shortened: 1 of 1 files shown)


kevincoleman:
Bump.

Ran into this little confusion; googled, it, and found this whole issue/PR thread. Simple change, just merge!

ljharb:
(07d1b473 added the "received" type)

Issue #28, opened 2017-04-13
Add ability to throw error instead of warning in tests


hakunin:
In tests it would be nice to throw an error if the property isn't of its expected type.

It would be nice if this could be scoped to our own components and not 3rd party ones using 3rd party components, but thats not necesary. 

People are already turning `console.warning` calls to throw errors, so this would be a nice upgrade for them.

iainbeeston:
This would be really useful for anyone using create-react-app, as PropType errors would appear in the new [error overlay](https://github.com/facebookincubator/create-react-app/pull/1101)

seanadenise:
Related to previous conversation [here](https://github.com/facebook/react/issues/1753).

uniphil:
I just published a [`check-prop-types`](https://github.com/ratehub/check-prop-types) package that does this. The main export returns the error instead of logging it, and a helper `assertPropTypes` throws it.

I'd love to submit changes to this library instead, but it seems like PRs here are not getting any attention.

artin-phares:
I'm trying to use `prop-types` to validate js schema of external service response on server.  
And surprised it's just generating console warnings.  
How do I supposed to react to validation errors?

`checkPropTypes` should definitely generate some report (like suggested in #88), or at least return true/false. Anyway it would be nice for user to decide how validation results should be handled (log, throw error, etc).

rumkin:
@artin-phares I've created [typed-props](https://github.com/rumkin/typed-props) library which has the same interface and produce an array of [validation report](https://github.com/rumkin/validation-report) issues.

franciscop-invast:
I am doing this right now, and it seems to work pretty sweet:

```js
describe("...", () => {
  beforeAll(() => {
    jest.spyOn(console, "error").mockImplementation(error => {
      throw error;
    });
  });

  afterAll(() => {
    console.error.mockRestore();
  });

  // ...
});
```

It will, only for testing, convert every `console.error(error)` into a `throw error`. Something similar could be done for warnings.

a-ignatov-parc:
`nextTick` will be called before the current event loop continues, so it will throw after the warning is printed and before everything else.

rumkin:
@a-ignatov-parc Thrown error will loose the stack in that way and it will not be captured in the place where it was thrown. Also printing of PropTypes' message will not be stopped.

sarbbottam:
I am not sure what is the current state of this issue and the [project](https://github.com/facebook/prop-types/issues/249).

[...]

PR #54, opened 2017-05-17
checkPropTypes: add flag to throw the error instead of logging

### LINKED ISSUE #34: Using prop-types outside of React
Now that `prop-types` is a top-level package, it occurs to me that there are several use cases where this could be a useful library outside of React.

## Use Case

Suppose, for example, you are writing an application with a REST interface that takes in some complicated JSON data, and queries MongoDB:

```js
app.post("/api/getLastUserLogin", (req, res) => {
    // Expect req.body to be a `{username}` object.
    db.users.find({username: body.username})
    .then(users => {
        res.json(users[0] && users[0].lastLoginTime);
    });
});
```

Simple enough... But, oh noes!  There are dark forces at work on the internet, and someone sends us this JSON body:

```js
{"username": {"$ne": ""}}
```

MongoDB faithfully tries to load all 10 million of our users into memory, the server crashes, our company goes bankrupt, and our children starve.

But wait!  `prop-types` to the rescue!

```js
const LastUserLoginMessageType = {
    username: PropTypes.string.isRequired
}

app.post("/api/getLastUserLogin", (req, res) => {
    if(!PropTypes.validate(LastUserLoginMessageType, req.body)) {
        return res.status(400).send("Not today, hax0r!");
    }
    ...
});
```

## Why this is awesome

There's a lot to recommend this.  First, it's easy to check types without having to pull in some extra build step like TypeScript or Flow (especially nice on existing projects which don't already incorporate those technologies).  Second, it's a syntax that many people are already very familiar with, because it's so widely used in React.

The challenge is, of course, that today prop-types is so tightly coupled to React.  For starters, all of prop-types is disabled if `NODE_ENV === 'production'`, which is an obvious problem, so there'd need to be some changes.  Turning prop-types into a more general-purpose library should have advantages for React though, as it should make it easier to do things like https://github.com/reactjs/prop-types/issues/28.

## API Proposal

This is just to give a vague idea of what I'm after here, there's lots of room for improvement on this, but:

`PropTypes.isInvalid(typeSpec, value, options)`

* `typeSpec` is an object where keys are property names, and values are the associated PropType objects (e.g. `{username: PropTypes.string}`).
* `value` is the object we're going to check.
* `options.context` - See `options.onValidationError()`.
* `options.onValidationError({path, expectedType, actualType, errorString, context})` - If this function returns a truthy value, then `isInvalid()` will return that value immediately.  If this function throws, the the error will propagate all the way to the caller of `isInvalid()`.  `path` is the path of the value being checked (e.g. "user.name").  `expectedType` is the type that was expected at this path (e.g. 'string'), actualType is the type we actually found.  `context` is the `options.context` object passed to `isInvalid()`.  `errorString` is an english-language ready made error string to describe what went wrong.  The default implementation simply returns `errorString`.

I think this gives you flexibility to do quite a bit.  The existing React use case, where we need to know if `value` came from props or context or whatever is solved by passing `{location, componentName}` along in the `context`, so `onValidationError()` can construct an appropriate string.  React would set `onValidationError: () => null` when running production code.

One minor downside to this is it makes a webpacked production React app slightly bigger, since all this code that used to get optimized out is no longer going to be enclosed in `NODE_ENV === 'production'` checks.  I can think of a couple of ways around that, but I'll leave that for future discussion.

rufman:
When using prop-types for checking things outside of React, having `checkPropTypes` throw an error instead of swallowing it and logging it to `console.error` can be useful. The flag `throwErrors` will throw the error if set to true instead of logging it.  A`checkPropTypeWithErrors` convenience function, which calls `checkPropTypes` with `throwErrors` set to `true` has also been added.

Fixes #34


rufman:
@gaearon Is this a good approach to using prop-types checking outside of React? Or is there a better way?

gaearon:
Sorry, we're busy with many other things right now, and it's unlikely we'll have time to review PRs to `prop-types` very soon. In this PR, a few things concern me:

* Creating errors is expensive and we'd like to get away from that, but `throw`ing reinforces this pattern
* This doesn't expose other important information (such as component stack)

rufman:
The PR has been updated to not throw an error, but instead calls the function passed in as an argument, with the same arguments as are passed to `warning`.

rufman:
I'm aware that there might not be time to review this PR, but with React 16 in beta now the resolution on how to use prop-types with custom errors is becoming a more prominent issue. 

I was wondering if there is anything else that would need to be done on this PR to have it merged. Or if such a feature won't be considered for the prop-types package, we'll look into alternatives.

kaiyoma:
Any traction here? This is something we're interested in as well.

martinhorsky:
@kaiyoma I used `check-prop-types` module mentioned here: https://github.com/facebook/prop-types/issues/28#issuecomment-313504898

JoeDuncko:
This would also be super useful to me. Any word on whether this might make it in? If it won't, then closing this PR and "officially" referring users to another project that does this might be a good move.

checkPropTypes.js
```diff
@@ -25,7 +25,7 @@ if (process.env.NODE_ENV !== 'production') {
  * @param {?Function} getStack Returns the component stack.
  * @private
  */
-function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
+function checkPropTypes(typeSpecs, values, location, componentName, getStack, warningLogger = null, shouldLogAllTypeFailures = false) {
```
ljharb:
```suggestion
function checkPropTypes(typeSpecs, values, location, componentName, getStack, warningLogger = warning, shouldLogAllTypeFailures = false) {
```

ljharb: This boolean argument seems unnecessary to me, since you can encode that logic in your custom logger function.

kushthedude: @ljharb , its been a year since the PR is made , if you say i can make a new updated PR without any merge conflicts

ljharb: @kushthedude the time elapsed shouldn’t be relevant; please do *not* make a separate PR.

rufman: Removed the flag since it's now irrelevant (since https://github.com/facebook/prop-types/commit/40af72f41a404f081f87e5f160bb70c60401e587)

checkPropTypes.js
```diff
@@ -39,7 +39,9 @@ if (process.env.NODE_ENV !== 'production') {
  * @param {?Function} getStack Returns the component stack.
  * @private
  */
-function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
+function checkPropTypes(typeSpecs, values, location, componentName, getStack, warningLogger) {
+  var warningLogger = warningLogger || null;
```
ljharb: this line isn't necessary; if it's falsy, it will skip the `if` branch below already.
```suggestion
```

rufman: Fixed

jkusa:
Any update or ETA here?

cythrawll:
Man I sure could use this right now. Any updates?

JimmayVV:
oh my gosh so close, would love to see this added to the library, for sure.

JimmayVV:
This has been here for quite a while, and have read little to no objections to merging this PR aside from "time" constraints. This would be a wonderful addition to the propTypes library and I am genuinely confused as to what it would take to consider merging it.

ljharb (changes requested):
I've rebased this, and also changed the `warningLogger` to default to undefined, so it's not part of the function's length.

The only thing this is missing is similar tests in the other three test files.

PR #59, opened 2017-05-18
Add notNull modifier

### LINKED ISSUE #57: disallow null
When a type is optional it allows `undefined` or `null`. Unfortuantly prop-types eats the `null` and doesn't run any validators, so there doesn't seem to be a way to create a custom validator to disallow nulls.

lukescott:
Fixes #57

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,10 @@ MyComponent.propTypes = {
   // A value of any data type
   requiredAny: PropTypes.any.isRequired,
 
+  // You can use `notNull` like `isRequired`
+  // to allow undefined but not null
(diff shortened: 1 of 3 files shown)


jharris4:
I was just thinking about creating a PR for almost this exact same feature, except I was thinking of calling it `isDefined` instead of `notNull`.

Basically, I have a case where I want to allow `null` but disallow `undefined`.

lukescott:
Closing in favor of #90

jharris4:
@lukescott thanks for the initial work on this, hope you don't feel like I stepped on your toes making a similar PR ;-)

PR #63, opened 2017-05-24
Downgrade deprecation error to warning

flarnie:
**what is the change?:**
Inlines a version of 'warning' that uses `console.warn` instead of `console.error`, and then uses that for the deprecation warning.

**NOTE: I'm not sure that this will only warn in development, which is a problem. We don't seem to have a '__DEV__' env. variable set up here, so I'm not sure how to limit the warnings to development?**

**why make this change?:**
This is a particularly noisy warning that was firing in 3rd party libraries, which then causes it to show up in product code. Users then can't silence it easily without updating their library. This makes it slightly less painful, hopefully, but still painful enough to get folks to update their code.

**test plan:**
`yarn test` - updated some tests to check for this.

**issue:**
https://github.com/facebook/react/issues/9398

### DIFF
--- a/__tests__/PropTypesDevelopmentReact15.js
+++ b/__tests__/PropTypesDevelopmentReact15.js
@@ -100,11 +100,11 @@ function expectWarningInDevelopment(declaration, value) {
   for (var i = 0; i < 3; i++) {
     declaration(props, propName, componentName, 'prop');
   }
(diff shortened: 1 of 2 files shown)


gaearon:
We have a `process.env.NODE_ENV !== 'production'` check before calling the method so it should be good. It's CommonJS equivalent of `__DEV__`.

PR #78, opened 2017-06-12
'jsx' code annotation changed to 'js'

benwiley4000:
`js` is recognized by npmjs.com (`jsx` isn't) and there was no JSX here anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ commonly used with React components.
 Here is an example of using PropTypes with a React component, which also
 documents the different validators provided:
 
-```jsx
+```js
 import React from 'react';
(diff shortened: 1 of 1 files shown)


benwiley4000:
I signed it! Thanks

ljharb:
Thanks - can you file a bug on npm's markdown parser so it understands jsx, and link it here?

benwiley4000:
I can do that - and I also think this should be changed from jsx, since
it's not JSX. ;)

Unless I misunderstand what the jsx parser does.

ljharb:
:-) Totally agree

PR #80, opened 2017-06-15
Temporarily revert "Add exact type checker from react native"

gaearon:
Reverts facebook/prop-types#41

cc @aweary 

There were two issues that were missed during review:

### Prod version

See [this comment](https://github.com/facebook/prop-types/blob/8265af568f0cdcd998d0ca8b28f6fe472105f7ac/factoryWithTypeCheckers.js#L94-). Lists in these two files must be in sync. This is essential because as far as I can see `PropTypes.exact({}).isRequired` will throw in production code.

We need to automate this check because I agree it's not obvious.

### Object.assign

I don't think this repo is set up to polyfill Object.assign yet, so this will throw in older browsers.

### Way forward

@aweary Can you fix those two issues? Happy to take the PR back in after that.

For now I have to revert because I don't have time to fix forward and I'm worried we'll forget about this, and ship a broken version.


ljharb:
`object.assign` is a great non-auto-shimming polyfill if you need one :-)

gaearon:
We're using `object-assign` in React repo so would make sense for consistency.
We should really set up Babel here to match our transforms..

ljharb:
(you should use `object.assign` there too :-p)

Regardless of which module you use, you should use https://www.npmjs.com/package/babel-plugin-transform-replace-object-assign so that babel correctly transpiles object spread *and* `Object.assign` to use the module.

aweary:
@gaearon sorry about that, I wasn't aware of either of those conditions ☹️ I can fix, but not until next week unfortunately.

PR #90, opened 2017-07-13
Add isDefined and isNotNull parallels to isRequired

### LINKED ISSUE #57: disallow null

jharris4:
Adds the ability to do the following:

```
PropTypes.string.isDefined // allows null but not undefined
PropTypes.string.isNotNull // allows undefined but not null
```

closes #57 and duplicates some work from #59

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -150,7 +150,7 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
       var manualPropTypeCallCache = {};
       var manualPropTypeWarningCount = 0;
     }
-    function checkType(isRequired, props, propName, componentName, location, propFullName, secret) {
+    function checkType(allowNull, allowUndefined, props, propName, componentName, location, propFullName, secret) {
       componentName = componentName || ANONYMOUS;
       propFullName = propFullName || propName;
 
@@ -186,21 +186,19 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
           }
         }
       }
-      if (props[propName] == null) {
-        if (isRequired) {
-          if (props[propName] === null) {
-            return new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required ' + ('in `' + componentName + '`, but its value is `null`.'));
-          }
-          return new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required in ' + ('`' + componentName + '`, but its value is `undefined`.'));
-        }
-        return null;
+      if (props[propName] === null) {
+        return allowNull ? null : new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required ' + ('in `' + componentName + '`, but its value is `null`.'));
+      } else if (props[propName] == null) {
+        return allowUndefined ? null : new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required in ' + ('`' + componentName + '`, but its value is `undefined`.'));
       } else {
         return validate(props, propName, componentName, location, propFullName);
       }
     }
 
-    var chainedCheckType = checkType.bind(null, false);
-    chainedCheckType.isRequired = checkType.bind(null, true);
+    var chainedCheckType = checkType.bind(null, true, true);
+    chainedCheckType.isRequired = checkType.bind(null, false, false);
+    chainedCheckType.isDefined = checkType.bind(null, true, false);
+    chainedCheckType.isNotNull = checkType.bind(null, false, true);
 
     return chainedCheckType;
   }
--- a/__tests__/PropTypesDevelopmentReact15.js
+++ b/__tests__/PropTypesDevelopmentReact15.js
@@ -81,6 +81,40 @@ function typeCheckFailRequiredValues(declaration) {
   expect(message3).toContain(unspecifiedMsg);
 }
 
+function typeCheckFailDefinedValues(declaration) {
+  var unspecifiedMsg = 'The prop `testProp` is marked as required in ' +
+    '`testComponent`, but its value is \`undefined\`.';
+
+  var propTypes = { testProp: declaration };
+
+  // Required prop is undefined
+  var message2 = getPropTypeWarningMessage(
+    propTypes,
+    { testProp: undefined },
+    'testComponent',
+  );
+  expect(message2).toContain(unspecifiedMsg);
+
+  // Required prop is not a member of props object
+  var message3 = getPropTypeWarningMessage(propTypes, {}, 'testComponent');
+  expect(message3).toContain(unspecifiedMsg);
+}
+
+function typeCheckFailNotNullValues(declaration) {
+  var specifiedButIsNullMsg = 'The prop `testProp` is marked as required in ' +
+    '`testComponent`, but its value is `null`.';
+
+  var propTypes = { testProp: declaration };
+
+  // Required prop is null
+  var message1 = getPropTypeWarningMessage(
+    propTypes,
+    { testProp: null },
+    'testComponent',
+  );
+  expect(message1).toContain(specifiedButIsNullMsg);
+}
+
 function typeCheckPass(declaration, value) {
   const propTypes = {
     testProp: declaration,
@@ -222,6 +256,22 @@ describe('PropTypesDevelopmentReact15', () => {
       typeCheckFailRequiredValues(PropTypes.string.isRequired);
     });
 
+    it('should warn for missing defined values', () => {
+      typeCheckFailDefinedValues(PropTypes.string.isDefined);
+    });
+
+    it('should not warn for defined values when the value is null', () => {
+      typeCheckPass(PropTypes.string.isDefined, null);
+    });
+
+    it('should warn for missing non null values', () => {
+      typeCheckFailNotNullValues(PropTypes.string.isNotNull);
+    });
+
+    it('should not warn for null values when the value is undefined', () => {
+      typeCheckPass(PropTypes.string.isNotNull);
+    });
+
     it('should warn if called manually in development', () => {
       spyOn(console, 'error');
       expectWarningInDevelopment(PropTypes.array, /please/);
@@ -230,42 +280,84 @@ describe('PropTypesDevelopmentReact15', () => {
       expectWarningInDevelopment(PropTypes.array.isRequired, []);
       expectWarningInDevelopment(PropTypes.array.isRequired, null);
       expectWarningInDevelopment(PropTypes.array.isRequired, undefined);
+      expectWarningInDevelopment(PropTypes.array.isDefined, /please/);
+      expectWarningInDevelopment(PropTypes.array.isDefined, []);
+      expectWarningInDevelopment(PropTypes.array.isDefined, null);
+      expectWarningInDevelopment(PropTypes.array.isNotNull, /please/);
+      expectWarningInDevelopment(PropTypes.array.isNotNull, []);
+      expectWarningInDevelopment(PropTypes.array.isNotNull, undefined);
       expectWarningInDevelopment(PropTypes.bool, []);
       expectWarningInDevelopment(PropTypes.bool, true);
       expectWarningInDevelopment(PropTypes.bool.isRequired, []);
       expectWarningInDevelopment(PropTypes.bool.isRequired, true);
       expectWarningInDevelopment(PropTypes.bool.isRequired, null);
       expectWarningInDevelopment(PropTypes.bool.isRequired, undefined);
+      expectWarningInDevelopment(PropTypes.bool.isDefined, []);
+      expectWarningInDevelopment(PropTypes.bool.isDefined, true);
+      expectWarningInDevelopment(PropTypes.bool.isDefined, null);
+      expectWarningInDevelopment(PropTypes.bool.isNotNull, []);
+      expectWarningInDevelopment(PropTypes.bool.isNotNull, true);
+      expectWarningInDevelopment(PropTypes.bool.isNotNull, undefined);
       expectWarningInDevelopment(PropTypes.func, false);
       expectWarningInDevelopment(PropTypes.func, function() {});
       expectWarningInDevelopment(PropTypes.func.isRequired, false);
       expectWarningInDevelopment(PropTypes.func.isRequired, function() {});
       expectWarningInDevelopment(PropTypes.func.isRequired, null);
       expectWarningInDevelopment(PropTypes.func.isRequired, undefined);
+      expectWarningInDevelopment(PropTypes.func.isDefined, false);
+      expectWarningInDevelopment(PropTypes.func.isDefined, function () { });
+      expectWarningInDevelopment(PropTypes.func.isDefined, null);
+      expectWarningInDevelopment(PropTypes.func.isNotNull, false);
+      expectWarningInDevelopment(PropTypes.func.isNotNull, function () { });
+      expectWarningInDevelopment(PropTypes.func.isNotNull, undefined);
       expectWarningInDevelopment(PropTypes.number, function() {});
       expectWarningInDevelopment(PropTypes.number, 42);
       expectWarningInDevelopment(PropTypes.number.isRequired, function() {});
       expectWarningInDevelopment(PropTypes.number.isRequired, 42);
       expectWarningInDevelopment(PropTypes.number.isRequired, null);
       expectWarningInDevelopment(PropTypes.number.isRequired, undefined);
+      expectWarningInDevelopment(PropTypes.number.isDefined, function () { });
+      expectWarningInDevelopment(PropTypes.number.isDefined, 42);
+      expectWarningInDevelopment(PropTypes.number.isDefined, null);
+      expectWarningInDevelopment(PropTypes.number.isNotNull, function () { });
+      expectWarningInDevelopment(PropTypes.number.isNotNull, 42);
+      expectWarningInDevelopment(PropTypes.number.isNotNull, undefined);
       expectWarningInDevelopment(PropTypes.string, 0);
       expectWarningInDevelopment(PropTypes.string, 'foo');
       expectWarningInDevelopment(PropTypes.string.isRequired, 0);
       expectWarningInDevelopment(PropTypes.string.isRequired, 'foo');
       expectWarningInDevelopment(PropTypes.string.isRequired, null);
       expectWarningInDevelopment(PropTypes.string.isRequired, undefined);
+      expectWarningInDevelopment(PropTypes.string.isDefined, 0);
+      expectWarningInDevelopment(PropTypes.string.isDefined, 'foo');
+      expectWarningInDevelopment(PropTypes.string.isDefined, null);
+      expectWarningInDevelopment(PropTypes.string.isNotNull, 0);
+      expectWarningInDevelopment(PropTypes.string.isNotNull, 'foo');
+      expectWarningInDevelopment(PropTypes.string.isNotNull, undefined);
       expectWarningInDevelopment(PropTypes.symbol, 0);
       expectWarningInDevelopment(PropTypes.symbol, Symbol('Foo'));
       expectWarningInDevelopment(PropTypes.symbol.isRequired, 0);
       expectWarningInDevelopment(PropTypes.symbol.isRequired, Symbol('Foo'));
       expectWarningInDevelopment(PropTypes.symbol.isRequired, null);
       expectWarningInDevelopment(PropTypes.symbol.isRequired, undefined);
+      expectWarningInDevelopment(PropTypes.symbol.isDefined, 0);
+      expectWarningInDevelopment(PropTypes.symbol.isDefined, Symbol('Foo'));
+      expectWarningInDevelopment(PropTypes.symbol.isDefined, null);
+      expectWarningInDevelopment(PropTypes.symbol.isNotNull, 0);
+      expectWarningInDevelopment(PropTypes.symbol.isNotNull, Symbol('Foo'));
+      expectWarningInDevelopment(PropTypes.symbol.isNotNull, undefined);
       expectWarningInDevelopment(PropTypes.object, '');
       expectWarningInDevelopment(PropTypes.object, {foo: 'bar'});
       expectWarningInDevelopment(PropTypes.object.isRequired, '');
       expectWarningInDevelopment(PropTypes.object.isRequired, {foo: 'bar'});
       expectWarningInDevelopment(PropTypes.object.isRequired, null);
       expectWarningInDevelopment(PropTypes.object.isRequired, undefined);
+      expectWarningInDevelopment(PropTypes.object.isDefined, '');
+      expectWarningInDevelopment(PropTypes.object.isDefined, { foo: 'bar' });
+      expectWarningInDevelopment(PropTypes.object.isDefined, null);
+      expectWarningInDevelopment(PropTypes.object.isNotNull, '');
+      expectWarningInDevelopment(PropTypes.object.isNotNull, { foo: 'bar' });
+      expectWarningInDevelopment(PropTypes.object.isNotNull, undefined);
     });
   });
 
@@ -286,11 +378,23 @@ describe('PropTypesDevelopmentReact15', () => {
       typeCheckFailRequiredValues(PropTypes.any.isRequired);
     });
 
+    it('should warn for missing defined values', () => {
+      typeCheckFailDefinedValues(PropTypes.any.isDefined);
+    });
+
+    it('should warn for missing non null values', () => {
+      typeCheckFailNotNullValues(PropTypes.any.isNotNull);
+    });
+
     it('should warn if called manually in development', () => {
       spyOn(console, 'error');
       expectWarningInDevelopment(PropTypes.any, null);
       expectWarningInDevelopment(PropTypes.any.isRequired, null);
       expectWarningInDevelopment(PropTypes.any.isRequired, undefined);
+      expectWarningInDevelopment(PropTypes.any.isDefined, null);
+      expectWarningInDevelopment(PropTypes.any.isDefined, undefined);
+      expectWarningInDevelopment(PropTypes.any.isNotNull, null);
+      expectWarningInDevelopment(PropTypes.any.isNotNull, undefined);
     });
   });
 
@@ -382,6 +486,18 @@ describe('PropTypesDevelopmentReact15', () => {
       );
     });
 
+    it('should warn for missing defined values', () => {
+      typeCheckFailDefinedValues(
+        PropTypes.arrayOf(PropTypes.number).isDefined,
+      );
+    });
+
+    it('should warn for missing non null values', () => {
+      typeCheckFailNotNullValues(
+        PropTypes.arrayOf(PropTypes.number).isNotNull,
+      );
+    });
+
     it('should warn if called manually in development', () => {
       spyOn(console, 'error');
       expectWarningInDevelopment(PropTypes.arrayOf({foo: PropTypes.string}), {
@@ -402,6 +518,22 @@ describe('PropTypesDevelopmentReact15', () => {
       );
       expectWarningInDevelopment(
         PropTypes.arrayOf(PropTypes.number).isRequired,
+        undefined,
+      );
+      expectWarningInDevelopment(
+        PropTypes.arrayOf(PropTypes.number).isDefined,
+        null,
+      );
+      expectWarningInDevelopment(
+        PropTypes.arrayOf(PropTypes.number).isDefined,
+        undefined,
+      );
+      expectWarningInDevelopment(
+        PropTypes.arrayOf(PropTypes.number).isNotNull,
+        null,
+      );
+      expectWarningInDevelopment(
+        PropTypes.arrayOf(PropTypes.number).isNotNull,
         undefined,
       );
     });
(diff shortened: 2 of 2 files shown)


factoryWithTypeCheckers.js
```diff
@@ -186,21 +186,19 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
-        if (isRequired) {
-          if (props[propName] === null) {
-            return new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required ' + ('in `' + componentName + '`, but its value is `null`.'));
-          }
-          return new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required in ' + ('`' + componentName + '`, but its value is `undefined`.'));
-        }
-        return null;
+      if (props[propName] === null) {
+        return allowNull ? null : new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required ' + ('in `' + componentName + '`, but its value is `null`.'));
+      } else if (props[propName] == null) {
```
MatthewHerbst: I think you meant `else if (props[propName] == undefined)` here?

jharris4: `== null` is the same as `=== undefined` in this case.

I used `== null` because that's what the old code used, and it avoids referencing the `undefined` value directly.

I think it's a holdover best practice from older browsers that allow undefined to be assigned a different value:

https://stackoverflow.com/questions/8783510/how-dangerous-is-it-in-javascript-really-to-assume-undefined-is-not-overwritte

MatthewHerbst: Right, but my point is you have the exact same check twice. Either get rid of one or make them different.

```
if (props[propName] === null) {

} else if (props[propName] === null) {
  // This block will never execute, ever
}
```

MatthewHerbst: Oh sorry, I didn't notice the `==` instead of it being `===`

jharris4: no worries, I make the same mistake all the time. I even use a font that helps me see the difference: https://github.com/tonsky/FiraCode

ljharb: `== null` also matches both null and undefined, which is the convention in propTypes, so it's *not* the same as `=== undefined`, by design.

jharris4:
@aweary any chance of getting some feedback on this PR?

I'd love to see this merged & released if possible, since there isn't a good workaround for supporting `null` values without this PR...

ljharb:
@jharris4 `explicitNull` in https://npmjs.com/airbnb-prop-types is a pretty good workaround :-)

jharris4:
@ljharb thanks for the suggestion, I just had a quick look at the source and tests for `explicitNull`...

I'm curious, how would you suggest using `explicitNull` to handle something like `PropTypes.String.isDefined` (which from this PR would allow `"abc"` or `null` but NOT `123` or `undefined`)

ljharb:
Using `or` from airbnb-prop-types also, `or([explicitNull, PropTypes.string]).isRequired` should do it.

aweary:
@jharris4 I'll try and review soon, thanks for doing this btw!

th0th:
Any updates on this? I really don't want to introduce a whole new package for nullable required props.

aweary:
@th0th I'm currently the only person working on this repo, and it's not something I can prioritize right now, so I'm sorry that it's been rather slow. 

@ljharb if you could give this a review and let me know how it looks to you I would feel better about merging this soon.

ljharb:
I'm not really sure this is a good approach; I'd prefer adding something like `PropTypes.allOf`, and then adding `PropTypes.null` and `PropTypes.undefined`, and then `PropTypes.allOf(PropTypes.string, PropTypes.null)` etc.

jharris4:
@ljharb Things like `allOf` or `oneOf` could be nice additions to PropTypes validation but that's a more complicated feature that's way beyond the scope of this PR.

This PR encompasses the minimal change possible to give people the ability to distinguish between `null` vs `undefined`.

This PR is also a non breaking change and keeps the PropTypes package completely backwards compatible with previous versions.

With the current state of the code most of the change that I made in this PR would be required anyway, since there is currently no way to allow `null` values with `isRequired` at present:

```
if (props[propName] == null) {
  if (isRequired) {
    if (props[propName] === null) {
      return new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required ' + ('in `' + componentName + '`, but its value is `null`.'));
          }
```

IF it helps, I can give you a little backstory on why allowing `null` but NOT `undefined` is a useful feature.

I work on a project where a configuration is serialized to/from JSON and used to configure the appearance of a React component.

There are dozens of keys for things like this:

```
{
  "mainTitle": "Hello World",
  "subHeading": {
    "subTitle": "Another Title"
  }
}
```

ljharb:
`oneOf` already exists; I was suggesting `allOf` to complement it.

I realize it's out of scope of this PR; I'm suggesting that this PR's premise is flawed and it should be closed.

jharris4:
I'd argue that the premise of the original code to conflate `null` with `undefined` was flawed, and that this PR provides a simple, backwards compatible, non-breaking change, fully tested change to work around that flaw.

I get the sense that there are other users out there who could also benefit from this change to allow them to continue using `prop-types` with sometimes `null` prop values. But if nobody else chimes in we can close this PR.

It'd be unfortunate though, because without this PR `prop-types` can't meet all of my prop validation needs, and I'll have to look for another solution.

I wonder whether the core React team have mostly switched to using `Flow` for prop type validation (https://flow.org/en/docs/react/components/) and whether that'll be the most officially supported solution moving forward.

jharris4:
Also with regards to `oneOf`, without this PR, there would be no way to accomplish this:

```

**Comment by MatthewHerbst** [comment]
@ljharb I'm kind of in the same boat as @jharris4. I agree that your solution makes more sense long-term, but it also requires a major breaking change to how `isRequired` works. The proposed solution is pretty clean, and doesn't per-se break any existing code patterns. In fact, if I'm reading what you're thinking fully through, I would argue that you support removing `isRequired` all together in favor of `oneOf/allOf` with a `PropTypes.undefined` as well.

**Comment by ljharb** [comment]
I'm not sure why it requires a major change at all - I'm not proposing changing how `.isRequired` works whatsoever.

I'm suggesting solely two semver-minor changes: adding `PropTypes.allOf`, and `PropTypes.explicitNull`.

If you insist on wanting to exclude `null`, then a third semver-minor change: adding `PropTypes.noneOf` as well.

From those primitives, everything this PR wants can be achieved, with composition.

**Comment by jharris4** [comment]
The trouble is that the current implementation of `isRequired` will not let you do any validation whatsoever if the prop value is `null` or `undefined`.

I have dozens of props in some of my apps that I'd like to validate to ensure that their values are always a `string` or `null` but never `undefined`.

The composition idea is a good one, so I'll run with it for a minute.

```PropTypes.allOf([PropTypes.not.undefined, PropTypes.String])```

Assuming that PropTypes supported negation (which it doesn't), I can't quite get what I want using `allOf` (correct me if I'm wrong).

```PropTypes.noneOf([PropTypes.undefined, PropTypes.not.String])```

Similarly with `noneOf`, I'm not quite seeing how to compose things together to get what I want.

```PropTypes.oneOf([PropTypes.null, PropTypes.String])```

This seems like the closest composition solution to what I actually want. The trouble is, it allows `undefined`. But then if I add `.isRequired` to it, it won't allow `null`. (i.e. this was the whole motivation for this PR lol).

**Comment by ljharb** [comment]
`PropTypes.oneOf([PropTypes.null, PropTypes.string.isRequired])`

**Comment by jharris4** [comment]
That could work, so long as `PropTypes.null` exists.

I would perhaps generalize it a bit to `PropTypes.value` so that you can do `PropTypes.value(null)` or whatever else you want.

**Comment by jharris4** [comment]
I just became aware of a detailed issue filed a while back on React that discusses a lot of these points! https://github.com/facebook/react/issues/2166

It's pretty funny to read the comment by @sebmarkbage:

```
I agree that null should be a valid value even if it's required. I don't think that the breaking change is too bad since it's more lenient. It wouldn't spam you. We could introduce nullable types PropTypes.Nullable(PropTypes.string).isRequired.

We could also rename isRequired -> isDefined
```

**Comment by ljharb** [comment]
OK, so, what about the following:
 - add the following semver-minor types:
   - `PropTypes.allOf(types)`
   - `PropTypes.noneOf(types)`
   - `PropTypes.value(v)`

That seems like it would address the use case.

Totally separately, it seems like #110 might be resolved by indeed adding `.isDefined` to all validators, but I think that's a larger change than adding new composable top-level types.

**Comment by jharris4** [comment]
I could see some value for `PropTypes.value` but don't have a strong opinion on the others, apart from a vague feeling that some sort of negation operator could probably reduce the need for them since `oneOf` already exists. (ex: `allOf(A, B) === !oneOf(!A, !B)`)

I guess I'm also struggling to see how adding completely new composite and value variants of PropType handles is not a significantly larger change than the `.isDefined` addition I've already made. Not to mention all the extra unit test cases that would be required...

Also to be blunt, I've already spent more time on this than I should have! (And thanks for baring with me this far lol)

The goal was to just fill the gaps in the PropTypes validation that exist surrounding `null` values, because I don't currently have time to look at migrating to another solution like `flow` (seems like that's where the React core team's focus is now: https://github.com/facebook/react/issues/2166#issuecomment-169143702)

For a lack of time, I may just end up making a fork of this project and using that to move forward, but it would be nice to have it merged into the official package and potentially even help out others who are facing the same obstacles...

**Comment by ljharb** [comment]
You can't negate `oneOf` using built-in propTypes, which is why you'd need `noneOf`.

`.isDefined` adds an extra property to existing types, which might break if someone is reflecting on them.

Flow can't fully replace PropTypes anyways, so I'm very interested in PropTypes continuing to evolve :-)

**Comment by jharris4** [comment]
Right, there'd need to be a new `PropTypes.not` added to support that.

I can't even visualize a case where someone would be doing that kind of reflection on PropType validator instances. Do you have an example of how/where somebody is doing that?

I'd also like to see `PropTypes` continue to evolve, but it seems a bit stuck unfortunately. The idea of adding `isDefined` was floated by members of the core React team almost 3 years ago, yet here we are...

So, I guess if we're at the end of the road for `.isDefined` I should just close this PR?

**Comment by MatthewHerbst** [comment]
I'm confused @ljharb . You say adding `isDefined` is a big change (that you seem to support), but this commit does it pretty trivially... Why are we so worried about a breaking change for a feature we all seem to agree on is a good thing (`isDefined`)? Bump the major version, explain the change in the CHANGELOG, update the README, and be done with it.

**Comment by ljharb** [comment]
This PR also adds `isNotNull`.

Per https://github.com/facebook/prop-types/pull/90#issuecomment-330943004, I do think `.isDefined` is a pretty solid solution to #110.

I don't think that `.isNotNull` is a good solution to "checking for or against null".

**Comment by jharris4** [comment]
I don’t personally plan to use `isNotNull` but I added it for completeness and because that’s what a prior PR here added.

The way I see it, if `isDefined` is valuable then the same holds true for `isNotNull`. Some people likely have valid uses for it...

I see them as the extensions to the toolbox that let you work with not just `!= null` (isRequired) but `!== null` (isNotNull) and `!== undefined` (isDefined) too.

If the name is one of the main yuck factors, I’m open to suggestions to replace `isNotNull`

**Comment by ljharb** [comment]
I think undefined is special, so `isDefined` is warranted, but I think the generic composable types I mentioned above are the better solution to handle `null`.

**Comment by jharris4** [comment]
@lukescott Care to chime in about the use cases that led you to create #59 for the `notNull` modifier?

@ljharb I can see both sides of the coin here.

`undefined` could be regarded as "special" since the `defaultProps` kick in based on that value (https://github.com/facebook/react/blob/357925a84e34d17b96568e67d30baeb2a399da11/src/isomorphic/classic/element/ReactElement.js#L236).

But then again, if you're using `defaultProps`, you probably don't really need `.isRequired' so that's kind of a moot point in my opinion.

It's more about building a foundation for the future, such that ALL PropTypes do not need to be prevented from performing any validation when the prop value is `null` OR `undefined`.

As things stand today all prop validation related to `null` or `undefined` has exactly 2 options:

a) don't use `.isRequired` and accept that any `null` or `undefined` values will never raise warnings no matter what.
b) use `.isRequired` and accept that none of your values can ever be `null` or `undefined`.

Adding `.isDefined` happens to give me what I need (the ability to allow `null`), but the lack of `.isNotNull` would still mean that nobody can allow `undefined`.

Why block one and not the other? People may have valid use cases for that. Also, JavaScript provides them as two distinct values, so why not allow them to be handled independently?

**Comment by jharris4** [comment]
@lukescott Thanks for the interesting take on why `isNotNull` is useful. It makes sense to me as well given that many of my prop values are coming from JSON which is what motivated me to use `null` instead of `undefined`...

The idea of deriving the prop types from TypeScript / Flow would be super interesting

There's actually a babel plugin for flow that looks interesting: https://github.com/thejameskyle/babel-plugin-react-flow-props-to-prop-types

Out of curiosity I just looked at how that plugin handles `null` / `undefined`, and it looks like it runs right smack into the issue this PR is trying to resolve:

Flow Type Input:
```
props: {
  a: null
};
```

PropType Output:
```
static propTypes = {
  a: _PropTypes.oneOf([null]).isRequired
};
```
(From: https://github.com/thejameskyle/babel-plugin-react-flow-props-to-prop-types/blob/master/src/__tests__/classes.test.js#L128)

I'd file an issue if they weren't disabled on that repo. lol

**Comment by gneu77** [comment]
Is there any progress on this?
I have more and more components, where for now (until there is a solution) I removed the .isRequired to get rid of the warnings on intentional null.
I'm pretty much in favor of .isDefined, because it's a simple non-breaking change (I think .isRequired should have allowed for null from the beginning, but I agree that for backward compatibility, this should not be changed). However, I would also go with a more complex syntax, as long as I can make my application to check for the right things (In my case, mostly presence of a certain key with a nullable string as value).
@jharris4 You mentioned you might be willing to fork this project. I'd be happy to use this fork, so please let me know in case you go for this solution.

**Comment by jharris4** [comment]
@gneu77 I forked this a long time ago but haven't updated recently.

https://www.npmjs.com/package/prop-types-defined
and
https://github.com/jharris4/prop-types

I should probably update it to match the latest from this repo (mostly just the patent license change I think)

**Comment by gneu77** [comment]
@jharris4 Ah, thanks for the links 👍

**Comment by kheyse-oqton** [comment]
I do hope this gets merged in.
Any chance that perhaps just the .isDefined can be merged already, since it seems there's consensus about this?

**Comment by jharris4** [comment]
I don't use `isNotNull` myself, but I believe it should be supported as well.

It really is unfortunate that this PR has been stuck for so long, mostly because of an original (short-sighted) decision in the core logic of `prop-types` to make it impossible to do any validation specific to `null` vs `undefined` values, and an apparent reluctance to rectify that.

A validation library should not be dictating what valid props can be, but instead should be providing the tools for developers to decide what is valid for their use cases.

This PR addresses that issue, but it's hard to see it moving forward unless more people affirm their support it... :-(

**Comment by barrystaes** [comment]
Theres lots of places I just want to allow `null` as a value, but not `undefined` which would indicate wrong usage.  I think the suggested `.isDefined` is a perfect solution for this.

**Comment by pushitreal** [comment]
Copying my comment from https://github.com/facebook/react/issues/3163:

> Requiring that a value or null must be specifically provided should absolutely be allowed via PropTypes.
>
> Here is the use case I'm running into. We have some callbacks that are required by 90% of our selectors. The few that don't need them are very specific use cases. We have some new developers that are constantly forgetting to provide all of the usual callbacks.
>
> I want to force all uses of these Components to make the conscious decision to not include specific callbacks, rather than someone just forgetting a few props.
>
> Yes, we can hack in our own more specific checks via flow, but that splits our props validation into two places, and is rather unintuitive for someone glancing at the propTypes definitions.

Specifically, the more intuitive syntax for me would be:
`PropTypes.oneOfType([
    PropTypes.func,
    PropTypes.null
]).isRequired`

**Comment by danielrob** [comment]
This is honestly ridiculous. One of the key reasons to use propTypes is to confirm that a prop is actually getting passed. If you can't allow it to be `null` and yet still have it as required, then... well there is no then. Go use another version of the library that allows this.

**Comment by KeitIG** [comment]
@danielrob Are you replying to @pushitreal or to this PR? this is really unclear. The purpose of this PR is completely valid (cf. the whole discussion here).

**Comment by danielrob** [comment]
@KeitIG sincere apologies it was with the thread. I am in complete and strong agreement with this PR.

**Comment by KeitIG** [comment]
There is no conflict with `.isRequired`, so I do not understand your point. `isDefined` here is basically `isRequiredOrNull`, this is quite common in React apps (as an initial state to make the difference between "empty" and "haven't fetched the data yet").

**Comment by adamchenwei** [comment]
can we merge this already? is there any concern that holding this up still?

**Comment by adamchenwei** [comment]
@ljharb would it possible to think the perfect solution in the future release and get this very much agreed upon solution into the code base? Lets, go "MVP" a bit for this piece of work and build the rocket later?

**Comment by ljharb** [comment]
It’s not up to me.

**Comment by jharris4** [comment]
Is there somebody specific we can make the case to then?

I've been using a forked package with this PR merged for ages now but I'd much rather just be able to use `prop-types`... ;-)

**Comment by aweary** [comment]
Hey everyone 👋 sorry for the delays, I know there's a lot of interest in this and other features. I'm currently working on figuring out what our long-term plans for this package are, and who are the "right" people to make the call here. I'll let you know what I find out 🙃

**Comment by jharris4** [comment]
I had a shower thought that `prop-types` might have been built the way it was to match the logic for how static `ReactElement.defaultProps` get applied.

The [documentation for defaultProps](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) doesn't say anything about how `props` are evaluated to know when to apply the `defaultProps` so I had a look into the code.

The source for how [ReactElement](https://github.com/facebook/react/blob/f88deda83bab316385f39e8479850527cda90607/packages/react/src/ReactElement.js#L223) applies defaults to props is pretty simple:

```
// Resolve default props
if (type && type.defaultProps) {
  const defaultProps = type.defaultProps;
  for (propName in defaultProps) {
    if (props[propName] === undefined) {
      props[propName] = defaultProps[propName];
    }
  }
}
```

Derived from this a simple setup to show how defaultProps differ for `null` vs `undefined` prop values is the following:

```
const theType = {
  defaultProps: {
    myProp: "123"
  }
};

function resolveDefaultProps(type, props) {
  // identical code to above
}
```

For `undefined` props the defaultProps work like this:

```
let theProps = {};

theProps: {
  myProp: void 0; // same as if myProp is not set
};
resolveDefaultProps(theType, theProps);

console.log(theProps); // logs { myProp: "123" }
```

For `null` props the defaultProps work like this:

```
theProps: {
  myProp: null
};
resolveDefaultProps(theType, theProps);

console.log(theProps); // logs { myProp: null }

```

TLDR React's built in `defaultProps` treat `null` and `undefined` differently.

There's a compelling argument that merging this PR would also bring the benefit of aligning `prop-types` more closely with `react` itself... ;-)

**Comment by TomYeoman** [comment]
Thanks for the PR isDefined sounds like a great addition.

I often set null as default initial error status and then change when necessary

**Comment by WattyRev** [comment]
Just changed jobs from working with Ember to working with React. [Ember Prop Types](https://github.com/ciena-blueplanet/ember-prop-types) already handles this by providing `PropTypes.null`. So, I could have a nullable value look like this:
```
myProperty: PropTypes.oneOfType([PropTypes.null, PropTypes.bool]).isRequired
```

**Comment by adamchenwei** [comment]
@aweary isn't the right people will be the ones who worked on this repo the most in facebook? I guess the reality maybe more complex than that? Would be interested to know the exact reason for delay though, just so everyone can grab their heads around why this is delayed sooo looong. But thanks for making this lib 👍

**Comment by CoVoCre** [comment]
This is so definitely needed since 3 years ago !!! Please merge, it makes us find detours every time although the fix is simple and straightforward !

**Comment by jeffvandyke** [comment]
Any updates after these few months? We still run into this issue in our code, and this seems to be one of those PRs that a lot of people definitely want to happen, and since this is the "blessed" repo by `create-react-app`, it seems important that this feature gap be fixed here.

**Comment by kylebebak** [comment]
I've read through most of the comments in this PR and related issues/PRs. The solution proposed by @jharris4 is excellent. It's backwards compatible (as many have already mentioned), with minimal changes to semantics, and is more intuitive than solutions proposed by @ljharb .

The workarounds are not serious. Use flow? Flow adds static type checking to all of JS. It's like using a sledgehammer to hang a picture. Flow isn't widely used in the React community. It seems more and more people are abandoning it in favor of TypeScript. Who knows where it'll be in a couple of years.

So why hasn't this been merged? As others have mentioned, it doesn't seem to be an issue of consensus. The reasonable explanations seem to be that the maintainers of this package don't care about it or they still haven't seen the PR.

It would be nice to get a straight answer. If that answer is "We don't care about this package, fuck off", that would be better than the current situation. At least people would know.

**Comment by darovic** [comment]
I would really like to see this merged as well, please. The length of time this has sat here is ridiculous. This PR was opened in July 2017 and it is now December 2018. @aweary

**Comment by jharris4** [comment]
I created an issue to add a deprecation/legacy notice to the PropTypes package since the core team at Facebook have said it's now considered legacy and they recommend using TypeScript instead of this package.

Here's the issue (with links to the tweets from the Facebook core team members if you're interested): #249

**Comment by patrickgaskill** [comment]
I just found this discussion after a junior dev ran into this very issue with nullable prop types. I'm not really interested in moving my team over to Typescript yet. Thank you for continuing to fight the good fight in here, @jharris4.

**Comment by Sbphillips19** [comment]
Will this ever be merged?  Seems like the best strategy I have seen so far for null props.

**Comment by adamchenwei** [comment]
@aweary been almost a year..... is this repo abandoned? If its so, plz just announce it officially, so everyone will start migrating to something else which will update the package more promptly (i.e. typescript).

**Comment by ljharb** [comment]
@adamchenwei no, not at all. there's been a number of releases in the last few months. This repo and package is still very much alive.

That doesn't - for any project - mean every PR will get merged, nor that maintainers will have time to review every PR.

**Comment by adamchenwei** [comment]
@ljharb I understand. thanks for the replying! This feature, however, I think is not just a small PR but a PR has some pretty significant impact. I understand when time discussion or consideration is needed, so we see progress in been made, though, in this case, this change seems have been in a good state but not getting enough attention, yet, the community in need it very much.
Do not get me wrong, I appreciate very very much for every one of your work in maintaining the library, just would like to see progress been made for an easily implemented and impactful feature to be dragged to death because of non-decision. I think everyone here can not understand is the fact that this setup did not receive any significant opposition, yet, it is still not getting pulled in. That may be a thing to address? Or disclose a proper reason in the significant delay (1 year +) regardless of the community welcome such change?

**Comment by ljharb** [comment]
Even adding a new top-level propType is a bit of a big deal, since this package gets so much usage - adding a new method to *every* top-level propType is much much riskier.

I've only been maintaining this package for a few months, and the limited time I've had to invest in it has been spent cleaning up bugfixes and cutting long-awaited releases - which naturally pushes risky additions to the bottom of the list, even if I'd personally enjoy using them.

I hope you, and everyone else, can be patient.

**Comment by jharris4** [comment]
Hey @ljharb, first of all thanks for donating your time to maintaining open source packages like this one. It can be a thankless job at times...

If you feel that there might be a way to move this PR forward, could you elaborate on the risks you're referring to?

The existing tests for this library all pass with this PR, and I took the trouble to add new tests to mitigate any risks. 

If you feel that more tests would be required to increase your confidence level, I could probably find some time to invest in that.

Also, in case it helps inspire confidence we've got 3 different products using this PR (via forked package) with a combined total of several hundred components, deployed on-premise to about 300 clients, and haven't had a single issue (we use [sentry](https://sentry.io) for error tracking).

**Comment by damianlajara** [comment]
How is this still open?! LGTM 🚀

**Comment by macandcheese** [comment]
Another vote to get this approved - how can we get this merged in? Having `isNotNull` to allow defaultProps to be set when `null` is returned would solve so many use graphQL et al use cases...

**Comment by macandcheese** [comment]
This is a great PR, should be merged. A consideration - does a failing PropType for either 
`isDefined` or `isNotNull` default to a provided DefaultProp? That would be 💯.

**Comment by icopp** [comment]
This change is still necessary to cover the simple use case of `InferProps` with Typescript and being able to pass through props to children that are strictly typed as `X | undefined`, thus rejecting `X | null | undefined` from `InferProps`. 

The current workaround looks like this (assuming `X` is `boolean`) and is both silly and verbose:

```
PropTypes.oneOfType([
  PropTypes.bool.isRequired,
  PropTypes.oneOf([undefined]).isRequired,
]).isRequired
```

**Comment by lukescott** [comment]
@icopp The workaround should't work because `null` and `undefined` are never passed to a validator. `isRequired` is a flag (not a validator) that changes the internal behavior of prop-types to emit an error rather than eat nullish values. If it was a validator we wouldn't need this PR and we could solve it ourselves.

The PR doesn't change prop-types behavior (for backwards compatibility), but it adds `isNotNull` and `isDefined` as companion flags to `isRequired` to treat `null` and `undefined` separately.

At this point is there a way to completely remove prop-types from React and use our own callbacks? If this package is no longer being maintained and the notion is to "use TypeScript" (which we do), should prop-types import just be removed from core and replaced with something a little more simplistic? Something that can be better extended?

On our team we really don't use prop-types at all anymore and use TypeScript, but I personally see the benefit of having runtime checking of props. The dream for me would be able to just define TypeScript types and have runtime checking in development to guard against badly typed endpoints.

**Comment by ljharb** [comment]
@lukescott this package is absolutely still being maintained - a new feature not being added, or taking a long time to be added, quite simply *never* means a package is unmaintained.

PropTypes and TypeScript can not replace each other; PropTypes is incapable of checking function argument types or return values; TypeScript is incapable of determining basic things like "is it an integer". They should be used together, in concert.

**Comment by adamchenwei** [comment]
Why can't this just be added as an "Experimental" feature?

**Comment by ljharb** [comment]
That's not a concept that exists in semver. If it's published, it's supported until at least the next major, and that's effectively the same as forever.

**Comment by adamchenwei** [comment]
I am more referring to some sort repos feature system in accepting features that community want but the owner of the repo do not want (for whatever reason) and see how this feature is been used and etc. Experimental features even in React library itself... Why can't it be also in prop-type repo? Just thought it can give some oxygen in this tight grip......

**Comment by seniath** [comment]
> This change is still necessary to cover the simple use case of `InferProps` with Typescript and being able to pass through props to children that are strictly typed as `X | undefined`, thus rejecting `X | null | undefined` from `InferProps`.

just chiming in to say i also ran into issues with this and `InferProps`, so it'd be great to get this merged in

**Comment by jdhorner** [comment]
I ended up here after realizing that I needed to support a prop that could be both null but also a value, but not undefined. I just realized that the PR has been open for over 3 years! I guess I'll keep my fingers crossed. :)

**Comment by ljharb** [comment]
@jdhorner in the meantime, https://npmjs.com/airbnb-prop-types has `explicitNull`.

**Comment by jharris4** [comment]
@jdhorner The `explicitNull` from `airbnb-prop-types` **only** allows `null` or `undefined` values ([shown by tests here](https://github.com/airbnb/prop-types/blob/master/test/explicitNull.jsx#L38)) so likely won't meet your requirements.

The fork of this package with this PR merged is available here if you'd like to use it: https://www.npmjs.com/package/prop-types-defined

**Comment by ljharb** [comment]
@jharris4 you'd use `explicitNull.isRequired` with `or` or `PropTypes.oneOf`, so it absolutely would meet everyone's requirements.

**Comment by jharris4** [comment]
@ljharb ah yes, combing `or` with `explicitNull` from `airbnb-prop-types` [would work](https://github.com/airbnb/prop-types/blob/master/test/or.jsx#L105).

So for anyone looking for a way to do what this PR is about you have two options:

```
import PropTypes from 'prop-types-defined';

const propTypes = {
  aBoolOrNullValue: PropTypes.bool.isDefined
};
```

or

```
import PropTypes from 'airbnb-prop-types';

const propTypes = {
  aBoolOrNullValue: PropTypes.or([PropTypes.bool.isRequired, PropTypes.explicitNull().isRequired]).isRequired
};
```

@ljharb any reason not to close this PR now? Seems like it's never going to be merged, and it might be better to just point people to another solution...

**Comment by kesupile-kesupile** [comment]
Is there any movement on this PR? It would be nice to be able to do the following:

```ts
PropTypes.oneOfType([PropTypes.string, PropTypes.oneOf([null])]).isRequired
```

As others have probably already mentioned, `null` and `undefined` are two different constructs and should not be treated as the same thing...

davidje13:
@kesupile-kesupile it's looking increasingly likely that this will never be merged.

A while ago I made a tiny plugin to allows nulls (it's a plugin rather than a fork like some of the other solutions so you don't need to worry about it going stale): https://www.npmjs.com/package/prop-types-nullable — it's tested, supports TypeScript, and the chosen syntax is compatible with common linter rules.

```sh
npm install --save prop-types-nullable
```

Usage:

```js
const PropTypes = require('prop-types');
const nullable = require('prop-types-nullable');
 
MyReactClass.propTypes = {
  foobar: nullable(PropTypes.func).isRequired,
};
```

(or just copy the 10-or-so lines of code for `nullable` into your own codebase if you don't want an extra dependency)

ljharb:
@davidje13 https://npmjs.com/airbnb-prop-types has had an `explicitNull` propType for years; i'd suggest using that since that package also has many other propType composition helpers.

As to whether this will be merged or not, it's a relatively impactful change, so it's unlikely to be merged *soon*, at least.

lukescott:
Also React no longer has prop-types as a dependency. Even though though you could use other libs before, it still included it, but not anymore. Better to use an alt lib like the Airbnb one or a fork of this one.

matuszeg:
Any updates on when/whether this will be reviewed? Its a small PR that offers a large amount of value. It would tremendously improve the project. All the other alternatives suggested are all very verbose. Seems odd to not move forward with this when it provides a simple solution with very little changes.

ljharb (changes requested):
I'm still not sure whether this is viable to land, but to be landable, it'd have to add similar tests to all four test files.

PR #99, opened 2017-08-10
Update README.md with improved importing for CDNs

virgofx:
Ensure both unpkg and cdnjs are listed as CDNs, deprecate existing alias CDN reference in preference for best practice explicit version, and updated wording similar to React documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,15 +20,31 @@ import PropTypes from 'prop-types'; // ES6
 var PropTypes = require('prop-types'); // ES5 with npm
 ```
 
-If you prefer a `<script>` tag, you can get it from `window.PropTypes` global:
+### CDN
 
+If you prefer to exclude `prop-types` from your application and use it 
+globally via `window.PropTypes`, the `prop-types` package provides 
+single-file distributions, which are hosted on the following CDNs:
+
+* [**cdnjs**](https://cdnjs.com/libraries/prop-types)
 ```html
 <!-- development version -->
-<script src="https://unpkg.com/prop-types/prop-types.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/prop-types/15.5.10/prop-types.js"></script>
 
 <!-- production version -->
-<script src="https://unpkg.com/prop-types/prop-types.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/prop-types/15.5.10/prop-types.min.js"></script>
 ```
+
+* [**unpkg**](https://unpkg.com/prop-types/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,16 +20,32 @@ import PropTypes from 'prop-types'; // ES6
...
+
+<!-- production version -->
+<script src="https://cdnjs.cloudflare.com/ajax/libs/prop-types/15.5.10/prop-types.min.js"></script>
+```
+
+* [**unpkg**](https://unpkg.com/prop-types/)
 ```html
 <!-- development version -->
-<script src="https://unpkg.com/prop-types/prop-types.js"></script>
+<script src="https://unpkg.com/prop-types@15.5.10/prop-types.js"></script>
```
ljharb: The actual best practice is "^15.5.10" - ie, auto-updating to non-majors.

virgofx: That's best practice for repository dependency management but not for CDNs. Unpkg does support capability to redirect with aliases and version ranges as mentioned in original comment. That's not a best practice as it results in an additional round trip redirect.

ljharb: Fair; it just means that every release will need to either update the readme, or go out of sync.

virgofx: Yes, valid point. It's the one downfall of using a specific revision.
React's documentation is out of sync with the latest... So are a bunch of other repositories that list this same "CDN" section in their Readme's.

If there's a more concise way to update the wording, I'm all for it. I pretty much just deferred to the React wording and am confident that anyone who is using a CDN will be smart enough to figure out how to automate that process into their build pipeline.

virgofx:
Ping @aweary

virgofx:
@ljharb Just following back. Are you okay with changes. Any other issues that would prevent this from being merged?

README.md
```diff
@@ -20,16 +20,32 @@ import PropTypes from 'prop-types'; // ES6
...
+* [**cdnjs**](https://cdnjs.com/libraries/prop-types)
+```html
+<!-- development version -->
+<script src="https://cdnjs.cloudflare.com/ajax/libs/prop-types/15.5.10/prop-types.js"></script>
+
+<!-- production version -->
+<script src="https://cdnjs.cloudflare.com/ajax/libs/prop-types/15.5.10/prop-types.min.js"></script>
+```
+
+* [**unpkg**](https://unpkg.com/prop-types/)
```
aweary: Can we reorder these so unpkg comes first? Since the React docs only list the unpkg links I think it's more users will already be using it.

virgofx:
Sure... will update.

PR #68, opened 2017-06-01
Enhance warning message when a PropType is not correct

miguelcg:
Closes https://github.com/facebook/react/issues/9068 that got already closed when prop-types was moved. There was another PR (https://github.com/facebook/react/pull/9140) for the same purpose that also got closed for the same reason.

CC @gaearon as he participated in https://github.com/facebook/react/pull/9140

Thanks.

PR #103, opened 2017-08-30
Improve invalid type warnings

aweary:
This is a follow up for #68 

* Move type warning utilities to `TypeWarning` module
* Use `getPostfixForTypeWarning` in the invalid validator check instead of `typeof`
* Improve copy in the invalid validator check
* Improve the type warning postfix by making sure strings are prefixed with "a" and putting backticks around literal values like `undefined` and `null`
* Remove redundant test for invalid validators and add checks for additional types


gaearon:
Sorry, never saw that :-( I'm not watching this repo.

PR #106, opened 2017-08-31
Give production PropTypes identity

benjeffery:
As part of code that renders a user entered string of HTML+JSX(ish) tags I'm using the PropTypes of components to cast attribute values to the appropriate type. E.g. `if (propType === PropTypes.number)                   value = Number(value);`
Since upgrading to this separate prop-types module this is failing as all `PropTypes` are the same `shim` in production and hence are all equal. 

To fix this I have made each `shim` a separate instance so that it has identity.

### DIFF
--- a/factoryWithThrowingShims.js
+++ b/factoryWithThrowingShims.js
@@ -14,38 +14,41 @@ var invariant = require('fbjs/lib/invariant');
 var ReactPropTypesSecret = require('./lib/ReactPropTypesSecret');
 
 module.exports = function() {
-  function shim(props, propName, componentName, location, propFullName, secret) {
-    if (secret === ReactPropTypesSecret) {
-      // It is still safe when called from React.
-      return;
-    }
-    invariant(
-      false,
-      'Calling PropTypes validators directly is not supported by the `prop-types` package. ' +
-      'Use PropTypes.checkPropTypes() to call them. ' +
-      'Read more at http://fb.me/use-check-prop-types'
-    );
-  };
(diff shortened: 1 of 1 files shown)


ljharb:
@aweary can this be reviewed/merged/released? @airbnb's use case is identical to the OP.

sophiebits:
Meh. I'd discourage people from trying to do their own logic based on prop types though. It'd be surprising to people who know React (since most components don't behave that way) and prevents optimizations that some people want to do, like stripping prop types from prod builds. I don't think merging this is that harmful either though.

gaearon:
I think we won't proceed with this. They're intentionally meant to be opaque, and if you need something more reliable it's better to use your own explicit mechanism.

benjeffery:
If this is your decision then a note needs to be made in the "typechecking-with-proptypes" documentation page, stating that prop types shouldn't be used in either component code or other logic.

This wasn't obvious to me. I was aware that React itself wasn't going to check them in production, but I didn't expect the prop types themselves to be unusable.

gaearon:
It is mentioned here: https://github.com/facebook/prop-types#difference-from-reactproptypes-dont-call-validator-functions. Maybe not in the detail you needed—feel free to send a PR!

Same goes for the section in the React docs (it documents the 15.x behavior so didn't include that part). I guess we can update it now.

AndyOGo:
May I ask to continue discussing in #231 
I just filled literally the same PR for same reason #232

PR #112, opened 2017-09-20
Add support for objects created with a null prototype in objectOf

getkey:
It's all in the title. :-)

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -12,6 +12,7 @@ var assign = require('object-assign');
 var ReactPropTypesSecret = require('./lib/ReactPropTypesSecret');
 var checkPropTypes = require('./checkPropTypes');
 
+var has = Function.call.bind(Object.prototype.hasOwnProperty);
 var printWarning = function() {};
 
 if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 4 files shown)


getkey:
I would be happy to add it but isn't it a bit overkill?

ljharb:
The maintainers may think so; but being robust is rarely overkill, especially when it's so easy.

factoryWithTypeCheckers.js
```diff
@@ -10,6 +10,7 @@
 var emptyFunction = require('fbjs/lib/emptyFunction');
 var invariant = require('fbjs/lib/invariant');
 var warning = require('fbjs/lib/warning');
+var has = require('has');
```
gaearon: Can we avoid taking on more npm deps please? Just inline it. Thanks!

gaearon:
(Sorry for contradictory advice. Since this package is very commonly used we're trying to trim the tree a little bit.)

factoryWithTypeCheckers.js
ljharb: In that case, `Function.call.bind(Object.prototype.hasOwnProperty)`

getkey:
No worries. :slightly_smiling_face:

gaearon:
Thanks

PR #130, opened 2017-10-21
Switch the build process to rollup.js

realityking:
Makes the minified UMD build 1.2kb smaller. Rollup also eliminates a lot of closures, I suspect the code would be a tiny bit faster.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,22 +34,24 @@
     "lint": "eslint .",
     "test": "npm run tests-only",
     "tests-only": "jest",
-    "umd": "NODE_ENV=development browserify index.js -t loose-envify --standalone PropTypes -o prop-types.js",
-    "umd-min": "NODE_ENV=production browserify index.js -t loose-envify -t uglifyify --standalone PropTypes  -p bundle-collapser/plugin -o | uglifyjs --compress unused,dead_code -o prop-types.min.js",
-    "build": "yarn umd && yarn umd-min",
+    "build": "rollup -c",
     "prepublish": "not-in-publish || yarn build"
   },
   "devDependencies": {
+    "babel-core": "6",
     "babel-jest": "^19.0.0",
     "babel-preset-react": "^6.24.1",
-    "browserify": "^16.5.0",
-    "bundle-collapser": "^1.3.0",
     "eslint": "^6.3.0",
     "in-publish": "^2.0.0",
     "jest": "^19.0.2",
+    "loose-envify": "^1.4.0",
     "react": "^15.5.4",
-    "uglifyify": "^5.0.2",
-    "uglifyjs": "^2.4.11"
+    "rollup": "^1.32.1",
+    "rollup-plugin-babel": "3",
+    "rollup-plugin-commonjs": "^9.3.4",
+    "rollup-plugin-node-resolve": "^2.1.1",
+    "rollup-plugin-replace": "^2.2.0",
+    "rollup-plugin-uglify": "^6.0.4"
   },
   "browserify": {
     "transform": [
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,48 @@
+import replace from 'rollup-plugin-replace';
+import nodeResolve from 'rollup-plugin-node-resolve';
+import commonjs from 'rollup-plugin-commonjs';
+import {uglify} from 'rollup-plugin-uglify';
+import babel from 'rollup-plugin-babel';
+
+function stripEnvVariables(env) {
+  return {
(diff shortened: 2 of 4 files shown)


realityking:
@aweary Could you have a look at this one?

realityking:
@aweary ping

gaearon:
Can you please rebase and I'll take a look?

realityking:
@gaearon Rebased! And thanks for taking care of fbjs in prop-types as well :)

gaearon:
While we're at it, could you also add Babel plugin for good measure?

realityking:
Done. Though I don't think it actually transforms anything 🤔

package.json
```diff
@@ -26,25 +26,24 @@
   "homepage": "https://facebook.github.io/react/",
   "dependencies": {
     "loose-envify": "^1.3.1",
-    "object-assign": "^4.1.1"
+    "object-assign": "^4.1.1",
+    "rollup-plugin-babel": "^3.0.4"
```
TrySound: Move this please to devDependencies

rollup.config.js
```diff
@@ -0,0 +1,47 @@
...
+function stripEnvVariables(production) {
+  return {
+    __DEV__: production ? 'false' : 'true',
+    'process.env.NODE_ENV': production ? "'production'" : "'development'",
+  };
+}
+
+export default [
+  createConfig('prop-types.js'),
+  createConfig('prop-types.min.js', {uglify: true, production: true})
```
TrySound: Instead of configurable config factory it's cleaner to just write two configs and reuse `input` and `output.name` like [here](https://github.com/renatorib/react-powerplug/blob/master/rollup.config.js#L23-L62).

rollup.config.js
```diff
@@ -0,0 +1,47 @@
...
+      name: 'PropTypes'
+    },
+    plugins: [
+      replace(stripEnvVariables(options.production)),
+      nodeResolve(),
+      commonjs(),
+      babel({
+        exclude: 'node_modules/**'
+      })
+    ]
```
TrySound: The order of plugins matter. commonjs and replace plugins are able to consume only stable spec syntax. Better to keep babel before others.

TrySound (changes requested):

realityking:
@TrySound Made the changes you suggested. Not sure if I like the new structure of the rollup config so I made that change in a different commit.

rollup.config.js
```diff
@@ -0,0 +1,48 @@
...
+    output: {
+      file: 'prop-types.js',
+      format: 'umd',
+      name: 'PropTypes'
+    },
+    plugins: [
+      babel({
+        exclude: 'node_modules/**'
+      }),
+      replace(stripEnvVariables(options.production)),
```
TrySound: false

rollup.config.js
```diff
@@ -0,0 +1,48 @@
...
+    output: {
+      file: 'prop-types.min.js',
+      format: 'umd',
+      name: 'PropTypes'
+    },
+    plugins: [
+      babel({
+        exclude: 'node_modules/**'
+      }),
+      replace(stripEnvVariables(options.production)),
```
TrySound: true

TrySound:
@gaearon I think the PR is ready to merge.

TrySound:
Hey @gaearon. Any thoughts on this?

TrySound:
Hi @gaearon! Do you have a time to take a look at this?

ljharb:
Due to bugs like https://github.com/rollup/rollup/issues/2540, I don't think rollup is safe to use on any project at this time. Thanks for the PR!

gaearon:
To be fair, we do use Rollup for React itself. Since prop-types doesn't do any feature detection I wouldn't worry about it.

realityking:
I‘m sorry, I‘m confused. As @gaeron pointed out, rollup is widely used for Reaxt and related libraries. Why wouldn‘t it be a good a fit for prop-types?

Rollup is not for every library and I wouldn‘t use it to bundle an application but it poses a clear win (size) for this lib.

ljharb:
@gaearon fair enough! reopening.

TrySound:
@gaearon Would be good to finally merge it.

gaearon:
@ljharb volunteered to help clean up the repo so I trust him to make these decisions.

realityking:
I went through my open PR and found this gem from almost 3 (!) years ago. I've rebased it and upgraded all added dependencies to be the same version as in the react repository. I'd still love to see it merged. What do y'all think?

As an aside, going to the newest version of the rollup dependencies would require upgrading babel to version 7. Happy to do that if there's interest in that.

PR #131, opened 2017-10-22
Include CI for master branch and PR's

joefraley:
In accordance with https://github.com/facebook/prop-types/issues/19#issuecomment-316482639,
includes config for TravisCI to run test reports on the master branch
and PR's to the master branch.

- Includes .travis.yml config file
- Includes current CI status badge in README

Closes #19.

Note: Travis was not building the yarn cache successfully until I rebuilt the lock file. `npm install` worked fine though.

The currently failing tests are fixed in https://github.com/facebook/prop-types/pull/129/files

PR #129, opened 2017-10-21
Fix failing tests

realityking:
Failure seems to have been introduced with #68, before then tests are passing.

On that note, it would be great if CircleCI were set up for this repo. This cost me quite a bit of time today.

PR #132, opened 2017-10-22
Fix warning message

joefraley:
Fixes test failures related to changes in warning messaging. These errors turned up during development of https://github.com/facebook/prop-types/pull/131.

```
FAIL __tests__/PropTypesDevelopmentStandalone-test.js (6.672s)
 ● PropTypesDevelopmentStandalone › checkPropTypes › should warn for invalid validators
 expect(received).toEqual(expected)
 Expected value to equal:
 "Warning: Failed prop type: testComponent: prop type `foo` is invalid; it must be a function, usually from the `prop-types` package."
 Received:
 "Warning: Failed prop type: testComponent: prop type `foo` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`."

 FAIL __tests__/PropTypesDevelopmentReact15.js (7.328s)
 ● PropTypesDevelopmentReact15 › checkPropTypes › should warn for invalid validators
 expect(received).toEqual(expected)
 Expected value to equal:
 "Warning: Failed prop type: testComponent: prop type `foo` is invalid; it must be a function, usually from the `prop-types` package."
 Received:
 "Warning: Failed prop type: testComponent: prop type `foo` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`."
```

Of course, these kinds of failures will occur from time to time as long as the warning strings are hard-coded into the test specs. The alternative is to symbolize them or create message factories. The downside would be that the specs become much less intuitive to read and write.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: node_js
+node_js: "node"
+cache:
+  directories:
+    - "node_modules"
+notifications:
+  email: false
+branches:
+  only:
+  - master
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# prop-types
+# prop-types [![Build Status](https://travis-ci.org/facebook/prop-types.svg?branch=master)](https://travis-ci.org/facebook/prop-types)
 
 Runtime type checking for React props and similar objects.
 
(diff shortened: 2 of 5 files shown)


joefraley:
Actually this is taken care of in https://github.com/facebook/prop-types/pull/129/files

PR #146, opened 2017-12-06
[Add ComponentOfType]

congwenma:
Adds `ComponentOfType` propType checker.

This helps check Component's type on the receiving end.
```js 
//  1. single expected
  propType = {
    label: elementOfType('label')
  }

// 2. multiple expected, multiple given
  propType = {
    children: elementOfType([ MyListItem, 'label' ])
  }

  propType = {
    children: elementOfType([ ...SVG_TAG_GROUP, MyCustomTag ])
  }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,8 @@ class MyComponent extends React.Component {
   }
 }
 
+const OtherComponent = () => <div/>
+
 MyComponent.propTypes = {
   // You can declare that a prop is a specific JS primitive. By default, these
   // are all optional.
@@ -93,11 +95,14 @@ MyComponent.propTypes = {
   // it as an enum.
   optionalEnum: PropTypes.oneOf(['News', 'Photos']),
 
+  optionalElementWithType: PropTypes.elementWithType(['div', OtherComponent]),
+
   // An object that could be one of many types
   optionalUnion: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.number,
-    PropTypes.instanceOf(Message)
+    PropTypes.instanceOf(Message),
+    PropTypes.elementWithType(['label'])
   ]),
 
   // An array of a certain type
--- a/__tests__/PropTypesDevelopmentReact15.js
+++ b/__tests__/PropTypesDevelopmentReact15.js
@@ -1044,6 +1044,100 @@ describe('PropTypesDevelopmentReact15', () => {
     });
   });
 
+  describe('ElementWithType Types', () => {
+    it('should fail for mismatched component.type', () => {
+      typeCheckFail(
+        PropTypes.elementWithType('div'),
+        <span/>,
+        "Warning: Failed prop type: Invalid prop `testProp` of type [span] supplied to `testComponent`, expected one of type `div`."
+      );
+    });
+
+    it('should pass for matched component.type', () => {
+      typeCheckPass(PropTypes.elementWithType('div'), <div />);
+    });
+
+    it('should pass if no component given', () => {
+      typeCheckPass(PropTypes.elementWithType('div'));
+    });
+
+    describe('factory error', () => {
+      it('should fail if using any value thats not a function nor a string for expectedTypes', () => {
+        spyOn(console, 'error');
+
+        PropTypes.elementWithType(1);
+
+        expect(console.error).toHaveBeenCalled();
+        expect(console.error.calls.argsFor(0)[0]).toContain(
+          "Invalid argument supplied to ElementWithType, expected an Html tag name or a Component."
+        );
+
+        typeCheckPass(PropTypes.elementWithType(1), <span />);
+      });
+    });
+
+    describe('with .isRequired', () => {
+      it('should fail if not passing in anything', () => {
+        typeCheckFail(
+          PropTypes.elementWithType('div').isRequired,
+          null,
+          "Warning: Failed prop type: The prop `testProp` is marked as required in `testComponent`, but its value is `null`."
+        );
+      });
+    });
+
+    describe('custom component', () => {
+      const MyComponent = () => <div></div>;
+      var myComponentPropType;
+      beforeEach(() => {
+        myComponentPropType = PropTypes.elementWithType(MyComponent)
+      });
+
+      it('should fail for mismatched component.type', () => {
+        typeCheckFail(
+          myComponentPropType,
+          <span />,
+          "Warning: Failed prop type: Invalid prop `testProp` of type [span] supplied to `testComponent`, expected one of type `MyComponent`."
+        );
+      });
+
+      it('should pass for matched component.type', () => {
+        typeCheckPass(myComponentPropType, <MyComponent />);
+      });
+    });
+
+    describe('given component mismatches with expected', () => {
+      it('should warn if given invalid element', () => {
+        [
+          '<div></div>',
+          { type: 'div' },
+          1,
+          ['div'],
+          'div'
+        ].forEach(invalidElement =>
+          typeCheckFail(
+            PropTypes.elementWithType('div'),
+            invalidElement,
+            "Warning: Failed prop type: Invalid prop `testProp` of component `testComponent` has been given invalid component."
+          )
+        );
+      });
+    });
+
+    describe('list of expected types', () => {
+      it('should warn if expected type does not include given component.type', () => {
+        typeCheckFail(
+          PropTypes.elementWithType('div'),
+          <span />,
+          "Warning: Failed prop type: Invalid prop `testProp` of type [span] supplied to `testComponent`, expected one of type `div`."
+        );
+      });
+      it('should pass if expected type include given component.type', () => {
+        typeCheckPass(PropTypes.elementWithType('div'), <div />)
+      });
+    });
+  });
+
   describe('Union Types', () => {
     it('should warn but not error for invalid argument', () => {
       spyOn(console, 'error');
--- a/__tests__/PropTypesDevelopmentStandalone-test.js
+++ b/__tests__/PropTypesDevelopmentStandalone-test.js
@@ -1118,6 +1118,100 @@ describe('PropTypesDevelopmentStandalone', () => {
     });
   });
 
+  describe('ElementWithType Types', () => {
+    it('should fail for mismatched component.type', () => {
+      typeCheckFail(
+        PropTypes.elementWithType('div'),
+        <span />,
+        "Warning: Failed prop type: Invalid prop `testProp` of type [span] supplied to `testComponent`, expected one of type `div`."
+      );
+    });
+
+    it('should pass for matched component.type', () => {
+      typeCheckPass(PropTypes.elementWithType('div'), <div />);
+    });
+
+    it('should pass if no component given', () => {
+      typeCheckPass(PropTypes.elementWithType('div'));
+    });
+
+    describe('factory error', () => {
+      it('should fail if using any value thats not a function nor a string for expectedTypes', () => {
+        spyOn(console, 'error');
+
+        PropTypes.elementWithType(1);
+
+        expect(console.error).toHaveBeenCalled();
+        expect(console.error.calls.argsFor(0)[0]).toContain(
+          "Invalid argument supplied to ElementWithType, expected an Html tag name or a Component."
+        );
+
+        typeCheckPass(PropTypes.elementWithType(1), <span />);
+      });
+    });
+
+    describe('with .isRequired', () => {
+      it('should fail if not passing in anything', () => {
+        typeCheckFail(
+          PropTypes.elementWithType('div').isRequired,
+          null,
+          "Warning: Failed prop type: The prop `testProp` is marked as required in `testComponent`, but its value is `null`."
+        );
+      });
+    });
+
+    describe('custom component', () => {
+      const MyComponent = () => <div></div>;
+      var myComponentPropType;
+      beforeEach(() => {
+        myComponentPropType = PropTypes.elementWithType(MyComponent)
+      });
+
+      it('should fail for mismatched component.type', () => {
+        typeCheckFail(
+          myComponentPropType,
+          <span />,
+          "Warning: Failed prop type: Invalid prop `testProp` of type [span] supplied to `testComponent`, expected one of type `MyComponent`."
+        );
+      });
+
+      it('should pass for matched component.type', () => {
+        typeCheckPass(myComponentPropType, <MyComponent />);
+      });
+    });
+
+    describe('given component mismatches with expected', () => {
+      it('should warn if given invalid element', () => {
+        [
+          '<div></div>',
+          { type: 'div' },
+          1,
+          ['div'],
+          'div'
+        ].forEach(invalidElement =>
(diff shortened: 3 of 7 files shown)


ljharb:
Technically this is an element, not a component - it's similar to `elementType` in https://www.npmjs.com/package/airbnb-prop-types.

README.md
```diff
@@ -89,11 +91,14 @@ MyComponent.propTypes = {
   // it as an enum.
   optionalEnum: PropTypes.oneOf(['News', 'Photos']),
 
+  optionalElementOfType: PropTypes.elementOfType(['div', OtherComponent]),
```
ljharb: why allow it to take an array of types instead of a single one? `PropTypes.oneOfType` is the established way to compose multiple validators.

Separately, I think "oneOf" is always in the name when it's taking an array of multiples; thus, I think `elementOfType` taking one, or `oneOfElementType` taking an array, is what would make sense - and I think the latter sounds silly.

factoryWithTypeCheckers.js
```diff
@@ -314,6 +315,36 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
 
+
+  function createElementOfTypeChecker(expectedTypes) {
+    var ACCEPTABLE_TYPES_OF_EXPECTED_TYPES = ['string', 'function'];
+    function getDisplayName (comp) {
+      return typeof comp === 'function' ? comp.displayName || comp.name : comp;
+    }
+
+    function validate(props, propName, componentName, location, propFullName) {
+      expectedTypes = [].concat(expectedTypes)
```
ljharb: This should throw if `expectedTypes` isn't an array (if it takes an array), like `oneOf`/`oneOfType` do.

README.md
congwenma: The specific problem I'm trying to solve is the oneOfType scenario, but considering the need to whitelist more than one type of components:
```js
<CustomTable>
  <CustomHeader/>
  <CustomTableRow/>
  <CustomTableRow/>
  <CustomTableRow/>
  <CustomFooter/>
</CustomTable>

// this is what I'm looking for
CustomTable.propTypes = {
  children: elementsOfType([CustomHeader, CustomTableRow, CustomFooter])
}

or

<SVGSection>
  <use/>
  <g></g>
  <rectangle/>
  <CustomSvgLine />
</SVGSection>

SVGSection.propTypes = {
  children: elementsOfType([ ...SVG_CHILDREN, ...MY_CUSTOM_SVG_CHILDREN ])
}
```

I'm hoping to capture these use cases so that when allow reusability of component, I'm always covered by this type checker.

factoryWithTypeCheckers.js
congwenma: i'm hoping to give user some flexibility in case they only want to pass a single expected element type.

README.md
ljharb: `PropTypes.oneOfType([ ...SVG_CHILDREN, ...MY_CUSTOM_SVG_CHILDREN ].map(x => PropType.elementOfType(x)))` works just fine, without overcomplicating the propType validator.

factoryWithTypeCheckers.js
ljharb: Flexibility in APIs leads to bugs; if they want to pass a single type, they can add `[ ]` without difficulty.

README.md
congwenma: @ljharb that sounds reasonable, I will make the change.

congwenma: @ljharb fyi, one feature we will lose from this change is a error message that says `expected one of type [div, span, CustomComponent]`, if we use `oneOfType`, we will only be seeing the generic message `Invalid prop supplied ...`

ljharb: I believe there’s a way oneOfType can read info off of the validators it receives to improve the error message, but I’m not sure what it is.

__tests__/PropTypesDevelopmentStandalone-test.js
```diff
@@ -871,6 +871,82 @@ describe('PropTypesDevelopmentStandalone', () => {
     });
   });
 
+  describe('ElementOfType Types', () => {
```
ljharb: Also, ‘of type” implies it receives propType validators; i think this should be called “elementOf”.

congwenma: we are really checking if the given object has a `.type` attribute that matches the given accepted, hows `ElementWithType`

README.md
congwenma: @ljharb i wonder how that would work if each proptype returns a different error message.

__tests__/PropTypesDevelopmentStandalone-test.js
ljharb: or maybe just `elementType`?

congwenma: sure.

ljharb: createElementTypeTypeChecker :-D

__tests__/PropTypesDevelopmentReact15.js
```diff
@@ -875,6 +875,89 @@ describe('PropTypesDevelopmentReact15', () => {
...
+        <span/>,
+        "Warning: Failed prop type: Invalid prop `testProp` of type [span] supplied to `testComponent`, expected one of type `div`."
+      );
+    });
+
+    it('should pass for matched component.type', () => {
+      typeCheckPass(PropTypes.elementWithType('div'), <div />);
+    });
+
+    it('should pass if no component given', () => {
```
ljharb: This needs a test case for the required version that fails

__tests__/PropTypesDevelopmentReact15.js
```diff
@@ -875,6 +875,89 @@ describe('PropTypesDevelopmentReact15', () => {
...
+      typeCheckPass(PropTypes.elementWithType('div'));
+    });
+
+    describe('factory error', () => {
+      it('should fail if using any value thats not a function nor a string for expectedTypes', () => {
```
ljharb: Wouldn’t this throw, synchronously?

__tests__/PropTypesDevelopmentReact15.js
```diff
@@ -875,6 +875,89 @@ describe('PropTypesDevelopmentReact15', () => {
...
+        );
+      });
+
+      it('should pass for matched component.type', () => {
+        typeCheckPass(myComponentPropType, <MyComponent />);
+      });
+    });
+
+    describe('given component mismatch', () => {
+      it('should warn if given component.type other than function or array of functions', () => {
```
ljharb: Array language can be removed here (also in the docs)

__tests__/PropTypesDevelopmentReact15.js
congwenma: what do you mean by that?

__tests__/PropTypesDevelopmentReact15.js
congwenma: going to change to `should warn if given invalid component`

congwenma:
- [ ] Squash
- [ ] Update other specs to match what `PropTypesDevelopmentReact15.js` covers.

__tests__/PropTypesDevelopmentReact15.js
congwenma: updated. I'm adding a task to cover the "syching specs" part.

__tests__/PropTypesDevelopmentReact15.js
ljharb: I mean that `PropTypes.elementWithType(x)` should throw, immediately, if `x` is not a string or a function - it shouldn’t have to wait until propType validations to do so.

congwenma: yes, updated factory spec.

factoryWithTypeCheckers.js
```diff
@@ -314,6 +315,36 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
     return createChainableTypeChecker(validate);
   }
 
+
+  function createElementWithTypeChecker(expectedType) {
+    var ACCEPTABLE_TYPES_OF_EXPECTED_TYPES = ['string', 'function'];
+
+    if (ACCEPTABLE_TYPES_OF_EXPECTED_TYPES.indexOf(typeof expectedType) === -1) {
+      process.env.NODE_ENV !== 'production' ? warning(false, 'Invalid argument supplied to ElementWithType, expected an Html tag name or a Component.') : void 0;
```
ljharb: this should not use `warning` here; I'm suggesting this have a literal `throw` statement.

congwenma: this would be different from the behaviors of other type checkers.

https://github.com/congwenma/prop-types/blob/add-componentOfType/factoryWithTypeCheckers.js#L277

ljharb: hm, that seems like a missed opportunity.

congwenma: wouldn't this be a useful pattern by `not making a big deal` about prop types in production?

ljharb: I'm not sure what you mean; invalid props are bugs. The point of not warning about them in production is performance; the exception I'm hoping for here would only apply to dev anyways, since the entire propType validator is compiled out anyways in production.

congwenma: sorry I thought you mean `replacing the entire line with a  throw`. 
I think what you are suggesting can make things easier and we'd be removing dependency on `fbjs/warning`, however I think theres some logic in `fbjs/warning` that helps debugging prop-types and handle some special cases, and this probably warrants a separate PR.

ljharb:
@congwenma would you mind rebasing this, and addressing outstanding comments?

congwenma:
> @congwenma would you mind rebasing this, and addressing outstanding comments?

Rebased! What still needs to be addressed?

ljharb:
@congwenma https://github.com/facebook/prop-types/pull/146#discussion_r155375373 in particular

factoryWithTypeCheckers.js
```diff
@@ -364,6 +365,36 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+      return emptyFunction.thatReturnsNull;
+    }
+
+    function getDisplayName (comp) {
```
ljharb:
```suggestion
    function getDisplayName(comp) {
```

factoryWithTypeCheckers.js
```diff
@@ -364,6 +365,36 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+    function validate(props, propName, componentName, location, propFullName) {
+      var propValues = [].concat(props[propName]);
+
+      if (propValues.some(pv => !isValidElement(pv))) {
+        return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` of component `' + componentName + '` has been given invalid component.');
+      }
+
+      var hasInvalid = propValues.some(propValue => expectedType !== propValue.type);
+      if (hasInvalid) {
+        return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` of type [' + propValues.map(pv => pv.type).map(getDisplayName).join(', ') + '] ' + ('supplied to `' + componentName + '`, expected one of type `' + getDisplayName(expectedType) + '`.'));
```
ljharb: we can't use arrow functions in this package, since it doesn't use babel.
```suggestion
        return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` of type [' + propValues.map(function (pv) return getDisplayName(pv.type)).join(', ') + '] ' + ('supplied to `' + componentName + '`, expected one of type `' + getDisplayName(expectedType) + '`.'));
```

factoryWithTypeCheckers.js
```diff
@@ -364,6 +365,36 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+    }
+
+    function getDisplayName (comp) {
+      return typeof comp === 'function' ? comp.displayName || comp.name : comp;
+    }
+
+    function validate(props, propName, componentName, location, propFullName) {
+      var propValues = [].concat(props[propName]);
+
+      if (propValues.some(pv => !isValidElement(pv))) {
```
ljharb:
```suggestion
      if (propValues.some(function (pv) { return !isValidElement(pv); })) {
```

factoryWithTypeCheckers.js
```diff
@@ -364,6 +365,36 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+      process.env.NODE_ENV !== 'production' ? warning(false, 'Invalid argument supplied to ElementWithType, expected an Html tag name or a Component.') : void 0;
+      return emptyFunction.thatReturnsNull;
+    }
+
+    function getDisplayName (comp) {
+      return typeof comp === 'function' ? comp.displayName || comp.name : comp;
+    }
+
+    function validate(props, propName, componentName, location, propFullName) {
+      var propValues = [].concat(props[propName]);
```
ljharb: this should just be `props[propName]`; there's no reason to support multiple values here.

factoryWithTypeCheckers.js
```diff
@@ -364,6 +365,36 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
     return createChainableTypeChecker(validate);
   }
 
+
+  function createElementWithTypeChecker(expectedType) {
+    var ACCEPTABLE_TYPES_OF_EXPECTED_TYPES = ['string', 'function'];
+
+    if (ACCEPTABLE_TYPES_OF_EXPECTED_TYPES.indexOf(typeof expectedType) === -1) {
```
ljharb: instead of making an array and using indexOf, this should repeat `typeof` here - it's actually optimized to be faster anyways. However, instead, why not:
```suggestion
    if (!isValidElementType(expectedType)) {
```
? (using `isValidElement` from `react-is`)?

factoryWithTypeCheckers.js
congwenma: Are you saying the support of multiple element types should be using `oneOf`?

ljharb: Yes.

factoryWithThrowingShims.js
```diff
@@ -51,6 +51,7 @@ module.exports = function() {
     objectOf: getShim,
     oneOf: getShim,
     oneOfType: getShim,
+    elementWithType: getShim,
```
ljharb: also let’s name this `elementType`

congwenma: elementType already exist, looks like this PR beated us to it. https://github.com/facebook/prop-types/commit/b67bbd40b5bf8b27475801ecdb7617071ceaccac. I am going to close my PR.

congwenma: elementType already exist, looks like this PR beated us to it. https://github.com/facebook/prop-types/commit/b67bbd40b5bf8b27475801ecdb7617071ceaccac.

ljharb: Whoops, indeed - looks like #211 may have replaced this.

ljharb:
Replaced with #211.

Thanks for the contribution!

PR #158, opened 2018-01-31
Add isDefined parallel to isRequired

kheyse-oqton:
Adds the ability to do the following:

```
PropTypes.string.isDefined // allows null but not undefined
```

Based on https://github.com/facebook/prop-types/pull/90 but removes

```
PropTypes.string.isNotNull // allows undefined but not null
```

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -150,7 +150,7 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
       var manualPropTypeCallCache = {};
       var manualPropTypeWarningCount = 0;
     }
-    function checkType(isRequired, props, propName, componentName, location, propFullName, secret) {
+    function checkType(allowNull, allowUndefined, props, propName, componentName, location, propFullName, secret) {
       componentName = componentName || ANONYMOUS;
       propFullName = propFullName || propName;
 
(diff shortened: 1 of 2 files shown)


factoryWithTypeCheckers.js
```diff
@@ -186,21 +186,18 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
-        if (isRequired) {
-          if (props[propName] === null) {
-            return new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required ' + ('in `' + componentName + '`, but its value is `null`.'));
-          }
-          return new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required in ' + ('`' + componentName + '`, but its value is `undefined`.'));
-        }
-        return null;
+      if (props[propName] === null) {
+        return allowNull ? null : new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required ' + ('in `' + componentName + '`, but its value is `null`.'));
+      } else if (props[propName] == null) {
```
ljharb: might as well only check for undefined here, since we already checked for null above?
```suggestion
      } else if (typeof props[propName] === 'undefined') {
```

ljharb:
@kheyse-oqton please sign the CLA when you get a chance.

PR #168, opened 2018-03-01
Fix type hinting for IDE for isRequired

pgilad:
isRequired is currently unrecognised by IDEs when using `arrayOf`, `shape` and others. This will fix the type-hinting for it.

See:

image


gaearon:
This seems like an IDE issue to me. Please report it to them instead.

The `isRequired` property is on the return value, not the function itself. Putting it on the function makes the API different between development and production which is something we can't do.

Let me know if I missed something!

PR #171, opened 2018-03-19
Updated all vars in test files to consts and lets.

barrymichaeldoyle:
It's better to define variables as consts and lets over vars.

### DIFF
--- a/__tests__/PropTypesDevelopmentReact15.js
+++ b/__tests__/PropTypesDevelopmentReact15.js
@@ -9,8 +9,8 @@
 
 'use strict';
 
-var React;
-var PropTypes;
+let React;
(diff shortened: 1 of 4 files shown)


barrymichaeldoyle:
Doesn't the babel preset sort that out?

ljharb:
In the test files, i suppose, but none of the production files in this repo get run through babel.

barrymichaeldoyle:
Ah right, didn't realize that. I'll update the PR to only update the tests.

barrymichaeldoyle:
There we go, I reset back to the commits that only affect the test files 👍

gaearon:
Seems fine, thanks. A PR adding a Babel step would also be great!

Issue #183, opened 2018-05-21
`objectOf` should not require `hasOwnProperty` to be a function


tor-ocean:
When using objects as dictionaries, and to avoid littering code with `hasOwnProperty` checks, it can be useful to use `object = Object.create(null)` instead of `object = {}` when creating the object.

Unfortunately, this causes the `PropTypes.objectOf` checker to throw:

```
Warning: Failed prop type: propValue.hasOwnProperty is not a function
```

as the code assumes that every value will have an `hasOwnProperty` method in its prototype chain:

https://github.com/facebook/prop-types/blob/01634d92e4e4cb5ffaf7de96ea8cb5da9b65e4f0/factoryWithTypeCheckers.js#L305

In my case, I will have to work around it by using a regular object and refactor the rest of my own code to do the same `hasOwnProperty` check.

I would still argue that this is a bug and needs to be fixed. When checking that a value is an object where every own key maps to a value of a certain type, it is self-contradictory that the code simultaneously requires that one specific key is going to map to a function. What was supposed to check that the value is a dictionary of a specific type now risks crashing the app when that is the case.

Imagine the case where the object is truly being used as a dictionary, even with a regular `{}` object, what's to stop the key `"hasOwnProperty"` to map to some other value? Malicious users might even cause a component to crash for other users by exploiting this.

ljharb:
This is a good fix; not because null objects are reasonable or because almost anyone would use them, but because `delete Object.prototype.hasOwnProperty` is a thing.

dferber90:
I am getting the same warning when defining this prop-type validation:

```js
PropTypes.shape({
  pathname: PropTypes.string.isRequired,
  query: PropTypes.objectOf(PropTypes.string),
}),
```

And when the passed value is:

```
{
  "pathname": "/foo",
  "query": {
    "perPage": "6",
    "sortBy": "createdAt",
    "sortDirection": "desc"
  }
}
```

It results in this error:

```
Warning: Failed prop type: propValue.hasOwnProperty is not a function
```

likerRr:
Any updates on it? I also faced the same issue, when validating object, created with `Object.create`

dferber90:
#187 is approved and solves this. It's waiting for a merge still.

PR #187, opened 2018-05-29
Support validation when hasOwnProperty is not in prototype

dferber90:
Avoids the `Warning: Failed prop type: propValue.hasOwnProperty is not a function` warning when `hasOwnProperty` is not defined in the prototype chain.

See #183 for details.

---

- On the original issue @ljharb [noted something about `delete Object.prototype.hasOwnProperty` being a thing](https://github.com/facebook/prop-types/issues/183#issuecomment-390845818). I did not quite get this comment, so this solution might not work because of that. Just close this PR in that case.
- I did not add tests for the combined case of `PropTypes.shape({ q: PropTypes.objectOf(PropTypes.string) })` I pointed out in https://github.com/facebook/prop-types/issues/183#issuecomment-392545102, as I didn't know where to place them

### DIFF
--- a/checkPropTypes.js
+++ b/checkPropTypes.js
@@ -12,6 +12,7 @@ var printWarning = function() {};
 if (process.env.NODE_ENV !== 'production') {
   var ReactPropTypesSecret = require('./lib/ReactPropTypesSecret');
   var loggedTypeFailures = {};
+  var has = Function.call.bind(Object.prototype.hasOwnProperty);
 
   printWarning = function(text) {
     var message = 'Warning: ' + text;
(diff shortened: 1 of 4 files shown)


checkPropTypes.js
```diff
@@ -28,7 +28,7 @@ if (process.env.NODE_ENV !== 'production') {
 function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
   if (process.env.NODE_ENV !== 'production') {
     for (var typeSpecName in typeSpecs) {
-      if (typeSpecs.hasOwnProperty(typeSpecName)) {
+      if (Object.prototype.hasOwnProperty.call(typeSpecs, typeSpecName)) {
```
ljharb: The next robustness step is to do `var has = Object.prototypr.hasOwnProperty` at module scope, and then here, do `has.call`.

dferber90:
I moved the declaration to module scope now (amended the commit) and added a test for objects with a `hasOwnProperty` property.

mhuggins:
Thanks for this fix!  Running into this as well.  Is there a reason it sits open?

dferber90:
@mhuggins It is ready to be merged from my side (aside from the merge conflicts now).

ljharb:
Rebased this; it's partially addressed by #112 but still includes tests and a fix.

mhuggins:
Thanks, guys! ❤️

PR #191, opened 2018-06-05
.set and .setOf

Pimm:
An implementation for `.set` and `.setOf`. Please hit me with your feedback, thoughts, and ideas.

I can implement `.map` and `.mapOf` in this PR as well.

@ljharb mentioned that `instanceof Set` does not work everywhere. Perhaps that has to be fixed.

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -110,6 +110,8 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
     oneOf: createEnumTypeChecker,
     oneOfType: createUnionTypeChecker,
     shape: createShapeTypeChecker,
+    set: createSetTypeChecker(),
+    setOf: createSetOfTypeChecker,
     exact: createStrictShapeTypeChecker,
   };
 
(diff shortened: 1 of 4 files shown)


factoryWithTypeCheckers.js
```diff
@@ -379,6 +381,52 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
     return createChainableTypeChecker(validate);
   }
 
+  function createSetTypeChecker(expectedType) {
+    function validate(props, propName, componentName, location, propFullName, secret) {
+      var propValue = props[propName];
+      if (!(propValue instanceof Set)) {
```
ljharb: indeed, this should module-level-cache the getter for Set.prototype.size, call-bind it, and invoke it here.

factoryWithTypeCheckers.js
```diff
@@ -379,6 +381,52 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+    return createChainableTypeChecker(validate);
+  }
+
+  function createSetOfTypeChecker(typeChecker) {
+    function validate(props, propName, componentName, location, propFullName) {
+      if (typeof typeChecker !== 'function') {
+        return new PropTypeError('Property `' + propFullName + '` of component `' + componentName + '` has invalid PropType notation inside setOf.');
+      }
+      var propValue = props[propName];
+      if (!(propValue instanceof Set)) {
```
ljharb: also here

factoryWithTypeCheckers.js
```diff
@@ -505,6 +553,8 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
         return 'date';
       } else if (propValue instanceof RegExp) {
         return 'regexp';
+      } else if (propValue instanceof Set) {
```
ljharb: and here

PR #194, opened 2018-06-19
Remove fbjs dependency

gaearon:
See reasons in https://github.com/facebook/react/pull/13069.
This PR is complementary.

I chose to inline a simplified version of `warning` in two files where we use it, and replacing invariant calls with `new Error`. In case somebody relies on `Invariant Violation` being the name (e.g. in tests), I inlined `.name` assignments to match the `invariant` logic.

Without this PR, `fbjs` would stay a transitive dep of React.

### DIFF
--- a/checkPropTypes.js
+++ b/checkPropTypes.js
@@ -7,11 +7,24 @@
 
 'use strict';
 
+var printWarning = function() {};
+
 if (process.env.NODE_ENV !== 'production') {
-  var invariant = require('fbjs/lib/invariant');
-  var warning = require('fbjs/lib/warning');
   var ReactPropTypesSecret = require('./lib/ReactPropTypesSecret');
   var loggedTypeFailures = {};
+
+  printWarning = function(text) {
+    var message = 'Warning: ' + text;
+    if (typeof console !== 'undefined') {
+      console.error(message);
+    }
+    try {
+      // --- Welcome to debugging React ---
+      // This error was thrown as a convenience so that you can use this stack
+      // to find the callsite that caused this warning to fire.
+      throw new Error(message);
+    } catch (x) {}
+  };
 }
 
 /**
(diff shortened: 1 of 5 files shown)


checkPropTypes.js
```diff
@@ -7,11 +7,24 @@
...
+  printWarning = function(text) {
+    var message = 'Warning: ' + text;
+    if (typeof console !== 'undefined') {
+      console.error(message);
+    }
+    try {
+      // --- Welcome to debugging React ---
+      // This error was thrown as a convenience so that you can use this stack
+      // to find the callsite that caused this warning to fire.
+      throw new Error(message);
```
ljharb: the intention here is for the debugger to pause on it?

gaearon: Yeah. (Not new code, just copy paste from fbjs/lib/warning)

checkPropTypes.js
```diff
@@ -36,20 +49,39 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
...
         }
-        warning(!error || error instanceof Error, '%s: type specification of %s `%s` is invalid; the type checker ' + 'function must return `null` or an `Error` but returned a %s. ' + 'You may have forgotten to pass an argument to the type checker ' + 'creator (arrayOf, instanceOf, objectOf, oneOf, oneOfType, and ' + 'shape all require an argument).', componentName || 'React class', location, typeSpecName, typeof error);
+        if (error && !(error instanceof Error)) {
+          printWarning(
+            (componentName || 'React class') + ': type specification of ' +
+            location + ' `' + typeSpecName + '` is invalid; the type checker ' +
+            'function must return `null` or an `Error` but returned a ' + typeof error + '. ' +
+            'You may have forgotten to pass an argument to the type checker ' +
+            'creator (arrayOf, instanceOf, objectOf, oneOf, oneOfType, and ' +
+            'shape all require an argument).'
```
bvaughn: 😅  Reading this kind of makes me wish you'd also copied over the format-args function too heh

factoryWithTypeCheckers.js
```diff
@@ -7,14 +7,32 @@
...
-var warning = require('fbjs/lib/warning');
 var assign = require('object-assign');
 
 var ReactPropTypesSecret = require('./lib/ReactPropTypesSecret');
 var checkPropTypes = require('./checkPropTypes');
 
+var printWarning = function() {};
+
+if (process.env.NODE_ENV !== 'production') {
+  printWarning = function(text) {
```
bvaughn: Why copy this in each file? No big deal, since it's small.

checkPropTypes.js
gaearon: I did at first, but Babel output was super ugly and I wasn't confident enough to write the same by hand (there was both some spreading and some resting) and not break something. And we can't use ES6 because we haven't set up a pipeline here 😛

factoryWithTypeCheckers.js
gaearon: I thought about putting into `prop-types/warning` but then figured someone will start using *that* as a "public" API. Might as well just tuck it away.

MariaDima:
Can someone please make a release, so that all modules that depend on prop-types can get the vulnerability fix?

gaearon:
1. This PR *was* released as `15.6.2`.
2. **There's no and has never been a vulnerability in `prop-types`.** Even if some transitive dependency of `fbjs` had a vulnerability at some point, `prop-types` only uses functions in `fbjs` that have no transitive dependencies.

ljharb:
@gaearon could you push all the git tags up for the releases? It seems that most of them (including the latest) are missing.

gaearon:
Pushed a tag for the latest release.

PR #204, opened 2018-08-09
Completely remove envify in favor of loose-envify

### LINKED ISSUE #203: Shouldn't `loose-envify` be in devDependencies instead of dependencies?
If I'm not mistaken, it's only used in the `umd` script in the package.json. If I understand correctly, this script builds a bundle for browsers, and will only be used in dev process. 

This would remove unused transitive dependencies when using a package that depends on prop-types.

autra:
Following the discussion in #203, this PR move `loose-envify` from dependencies to devDependencies. 

I also took the opportunity to choose `loose-envify` over `envify` everywhere, as they are doing the same thing anyway. I have no opinion whether `loose-envify` is better than `envify` or not. Using `yarn run test` does not show any measurable difference in term of execution time, so I can switch back to `envify` if you prefer.

Closes #203.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,13 +25,12 @@
   },
   "homepage": "https://facebook.github.io/react/",
   "dependencies": {
-    "loose-envify": "^1.3.1",
     "object-assign": "^4.1.1"
   },
   "scripts": {
     "test": "jest",
-    "umd": "NODE_ENV=development browserify index.js -t envify --standalone PropTypes -o prop-types.js",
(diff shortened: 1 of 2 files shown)


autra:
> We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file.

Oh yes. Should be ok now :-)

autra:
hi @ljharb, thanks for the review! Is there something I must do to get this merged?
Thanks!

ljharb:
We have to wait for a collaborator to review and merge it, as I’m not one.

ljharb:
`loose-envify`'s gotchas don't apply here, since we're not using env values in template literals, nor as anything but direct literal member access, so this seems fine.

autra:
Thanks!

Issue #200, opened 2018-07-31
Add forwardRef / valid component prop type


aldeed:
A component expecting another component to be passed in can do something like this:

```js
Button: PropTypes.oneOfType([PropTypes.string, PropTypes.func]).isRequired
```

However, if I pass in a `Button` component that has been wrapped in [forwardRef](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components), the prop type check fails. forwardRef returns an object with render function, not a function.

**Can there be a built-in prop type for the forwardRef object?**

**And/or maybe a prop type for "anything that works as a component", which would be func, forwardRef obj, and any strings recognized as built-in DOM components.**

See https://stackoverflow.com/questions/45315918/react-proptypes-component-class/45316411, which suggests:

```js
FancyButton: function (props, propName, componentName) {
    if(!props[propName] || typeof(props[propName].render) != 'function') {
      return new Error(`${propName}.render must be a function!`);
    }
  },
```

ljharb:
I think a `Component` propType would be better - that way as further types are created, it can be expanded to allow them (like forward refs)

aldeed:
Agree @ljharb. For reference, this is the custom prop type I made that is working:

```js
import { isValidElementType } from "react-is";

/**
 * We use an Error-like object for backward compatibility as people may call
 * PropTypes directly and inspect their output. However, we don't use real
 * Errors anymore. We don't inspect their stack anyway, and creating them
 * is prohibitively expensive if they are created too often, such as what
 * happens in oneOfType() for any type before the one that matched.
 */
function PropTypeError(message) {
  this.message = message;
  this.stack = "";
}
// Make `instanceof Error` still work for returned errors.
PropTypeError.prototype = Error.prototype;

function getComponentValidator() {
  function checkType(isRequired, props, propName, componentName, location, propFullName) {
    const val = props[propName];
    if (val == null) {
      if (isRequired) {
        if (props[propName] === null) {
          return new PropTypeError(`The ${location} '${propFullName}' is marked as required in '${componentName}', but its value is 'null'.`);
        }
        return new PropTypeError(`The ${location} '${propFullName}' is marked as required in '${componentName}', but its value is 'undefined'.`);
      }
    } else if (!isValidElementType(val)) {
      return new PropTypeError(`Invalid ${location} '${propFullName}' supplied to ${componentName}. Expected a string (for built-in components) or a class/function (for composite components).`);
    }
    return null;
  }

[...]

PR #211, opened 2018-09-09
Add a PropTypes for React elementType (ie. MyComponent)

### LINKED ISSUE #223: Proptype for Context.Provider
Is there a proptype for Context.Provider? If I use `element` I get the following error message:

```
Warning: Failed prop type: Invalid prop `ContextProvider` of type `object` supplied to `myComponent`, expected a single ReactElement.
```

and if I use `node` I get:

```
Warning: Failed prop type: Invalid prop `ContextProvider` supplied to `myComponent`, expected a ReactNode.
```

eXon:
There is a missing prop type that is quite common for a lot of people: `PropTypes.component`. The `PropTypes.element` already exists to support `<MyComponent myProp={<MyOtherComponent />} />`. However, there is no way to easily type check: `<MyComponent myProp={MyOtherComponent} />`. Before React 16.3, we were getting away with `PropTypes.func`. 

# ForwardRef in 16.3
However, with new features since React 16.3, there is a new challenge: forward ref! `forwardRef` is generating a React component, but in a different form: it is an object! This is breaking the old way of using the `PropTypes.func` and making the type checking unreliable. To fix this, we need a more complex prop type:

```
MyComponent.propTypes = {
  myProp: PropTypes.oneOf([
    PropTypes.func,
    PropTypes.shape({
      $$typeof: PropTypes.symbol
    })
  ]).isRequired
};
```

The problem with this:
1. Now we're leaking an internal implementation of React
1. If you pass an element (`<MyComponent />`) instead of a component (`MyComponent`), it would not fail. An addition check on the $$typeof value has to be made and now the complexity is getting out of hand for a prop type that should be simple.

To fix this, I've added `PropTypes.component` that would fix this problem by checking the `$$typeof` and only allow components.

# Higher-Order Component (HOC)
The reason why this prop type is so important to add is because of HOC. If you're using a library HOC, as soon as you add the decorator on your component using `forwardRef` (example react-cookie, see https://github.com/reactivestack/cookies/issues/172), your component is an object instead of a function and most libraries (example react-router) are using `PropTypes.func` to check if your passing a component. The typecheck wrongfully fails.

# What should be a component?
First we need to support regular stateless and stateful component. The best way to do that is by checking if it's a function. I don't think we can do better here unfortunately because of stateless components.

Then, there is special cases created by React like the forward ref. Here are all the `$$typeof` we can check and which one should be a component or not:

- REACT_PROVIDER_TYPE: Yes
- REACT_CONTEXT_TYPE: Yes
- REACT_FORWARD_REF_TYPE: Yes

- REACT_ELEMENT_TYPE: No
- REACT_PORTAL_TYPE: No
- REACT_FRAGMENT_TYPE: No
- REACT_STRICT_MODE_TYPE: No
- REACT_PROFILER_TYPE: No
- REACT_ASYNC_MODE_TYPE: No
- REACT_PLACEHOLDER_TYPE`: No

Forward ref is the most obvious use-case, but the context provider/consumer are technically component created by React. I just don't think it would be used much in a prop though.

# Remaining Concerns
Here are my main concerns with this PR:

## Bumping React version to 16 in devDependencies
The tests were running on the old 15 version of React, I'm guessing to make sure we don't break the old version. However, to test forward ref and the context API, we need at least React 16.3. I've updated the version, but an alternative would be to have two different package with the different version. Is it worth it? Also, should we bump the version of prop-types to 16? There is no breaking change, but the tests would be running on that version unless we run the tests on both by adding a proxy package pointing to React 15.

## Should we really call that new prop type component?
I think it's the best name, but in the previous code, the prop type element (and sometimes node) were referred as component. To me, a React component is `MyComponent` while a React element is `<MyComponent />`. Is it so for the community? Is the distinction clear enough?

**EDIT**: Been renamed `elementType` to follow `is-react` name.

## Where I've added `isValidComponent`
Looking at the `isValidElement` code, I tried to stay consistent with where to put the code. However, I'm not sure this is 100% clean. I can refactor this if you think this is a problem, but I've tried to stay consistent with the current state of the code.

## Should we have more validation?
We could add more validation. Is the component should be a specific one? Should the component have a specific set of prop type (not sure about the use cases)?

## Documentation
If this is merged, we would need to update the React documentation with the new prop type. I can help with that once this is merged.

Fixes #223.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,8 +78,11 @@ MyComponent.propTypes = {
   // (or fragment) containing these types.
   optionalNode: PropTypes.node,
 
-  // A React element.
+  // A React element (ie. <MyComponent />).
   optionalElement: PropTypes.element,
+
+  // A React element type (ie. MyComponent).
+  optionalElementType: PropTypes.elementType,
 
   // You can also declare that a prop is an instance of a class. This uses
   // JS's instanceof operator.
--- a/__tests__/PropTypesDevelopmentStandalone-test.js
+++ b/__tests__/PropTypesDevelopmentStandalone-test.js
@@ -18,6 +18,7 @@ function resetWarningCache() {
   React = require('react');
   PropTypes = require('../index');
 }
+resetWarningCache();
 
 function getPropTypeWarningMessage(propTypes, object, componentName) {
   if (!console.error.calls) {
@@ -552,6 +553,77 @@ describe('PropTypesDevelopmentStandalone', () => {
       expectThrowsInDevelopment(PropTypes.element, false);
       expectThrowsInDevelopment(PropTypes.element.isRequired, null);
       expectThrowsInDevelopment(PropTypes.element.isRequired, undefined);
+    });
+  });
+
+  describe('ElementType Types', () => {
+    it('should support native component', () => {
+      typeCheckPass(PropTypes.elementType, 'div');
+    });
+
+    it('should support stateless component', () => {
+      var MyComponent = () => <div />;
+      typeCheckPass(PropTypes.elementType, MyComponent);
+    });
+
+    it('should support stateful component', () => {
+      class MyComponent extends React.Component {
+        render() {
+          return <div />;
+        }
+      }
+      typeCheckPass(PropTypes.elementType, MyComponent);
+    });
+
+    (React.forwardRef ? it : it.skip)('should support forwardRef component', () => {
+      const MyForwardRef = React.forwardRef((props, ref) => <div ref={ref} />);
+      typeCheckPass(PropTypes.elementType, MyForwardRef);
+    });
+
+    (React.createContext ? it : it.skip)('should support context provider component', () => {
+      const MyContext = React.createContext('test');
+      typeCheckPass(PropTypes.elementType, MyContext.Provider);
+    });
+
+    (React.createContext ? it : it.skip)('should support context consumer component', () => {
+      const MyContext = React.createContext('test');
+      typeCheckPass(PropTypes.elementType, MyContext.Consumer);
+    });
+
+    it('should warn for invalid types', () => {
+      typeCheckFail(
+        PropTypes.elementType,
+        true,
+        'Invalid prop `testProp` of type `boolean` supplied to ' +
+          '`testComponent`, expected a single ReactElement type.',
+      );
+
+      typeCheckFail(
+        PropTypes.elementType,
+        {},
+        'Invalid prop `testProp` of type `object` supplied to ' +
+          '`testComponent`, expected a single ReactElement type.',
+      );
+
+      typeCheckFail(
+        PropTypes.elementType,
+        [],
+        'Invalid prop `testProp` of type `array` supplied to ' +
+          '`testComponent`, expected a single ReactElement type.',
+      );
+
+      it('should warn for missing required values', () => {
+        typeCheckFailRequiredValues(PropTypes.elementType.isRequired);
+      });
+    });
+
+    it('should warn for React element', () => {
+      typeCheckFail(
+        PropTypes.elementType,
+        <div />,
+        'Invalid prop `testProp` of type `object` supplied to ' +
+          '`testComponent`, expected a single ReactElement type.',
+      );
     });
   });
 
--- a/index.js
+++ b/index.js
@@ -6,21 +6,12 @@
  */
 
 if (process.env.NODE_ENV !== 'production') {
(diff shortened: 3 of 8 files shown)


index.js
```diff
@@ -11,16 +11,56 @@ if (process.env.NODE_ENV !== 'production') {
...
   var isValidElement = function(object) {
     return typeof object === 'object' &&
       object !== null &&
       object.$$typeof === REACT_ELEMENT_TYPE;
   };
 
+  var isValidComponent = function(component) {
+    // Stateless or stateful component
+    if (typeof component === 'function') {
+      return true;
```
ljharb: this seems like it would return a lot of false positives, but i don't know of a better way to cover SFCs. It seems like an early `return true` for class-based components, though, would be better. You may want to use the `react-is` package here, and check the various "is" methods contained within.

index.js
```diff
@@ -11,16 +11,56 @@ if (process.env.NODE_ENV !== 'production') {
     Symbol.for('react.element')) ||
     0xeac7;
 
+  var REACT_PROVIDER_TYPE =
```
ljharb: these should be pulled from react-is and not hardcoded here.

index.js
eXon: I agree, it was my first though but went for a first version before digging into this. If it's a function (not a class), it should accept it. If it's a class, we should check if it inherits `React.Component` or from `React.createClass`. Not sure if I can do this with enough reliability.

index.js
eXon: I would have to add the `react-is` as a production dependency, are you open to that? I saw you were already hardcoding the element type. I wouldn't mind switching all of them.

ljharb: i'm not a collaborator on this repo, but personally yes, i am - it's minified heavily in production, and you could also wrap the require in a NODE_ENV check.

eXon:
This is related to https://github.com/facebook/prop-types/issues/200. Would probably fix the issue.

README.md
```diff
@@ -81,6 +81,9 @@ MyComponent.propTypes = {
   // A React element.
   optionalElement: PropTypes.element,
 
+  // A React component.
+  optionalComponent: PropTypes.component,
```
ljharb: this needs to be renamed also?

eXon:
Turns out `is-react` have everything elementType needs. Even though we depend on the `16..5.0` version, I don't think it would have any problem to work on older versions including 15.

__tests__/PropTypesDevelopmentStandalone-test.js
```diff
@@ -486,6 +486,77 @@ describe('PropTypesDevelopmentStandalone', () => {
     });
   });
 
+  describe('ElementType Types', () => {
+    it('sould support native component', () => {
```
ljharb: `s/sould/should/g` (but it's even better when tests don't use the word "should" - ie, `it supports native components`)

taion:
BTW, we've had this implemented in prop-types-extra as `elementType` for a while. We've found it helpful to include an additional warning when the user accidentally passes in an element: https://github.com/react-bootstrap/prop-types-extra/blob/v1.1.0/src/elementType.js.

eedrah:
Closed issue #223 in favor of this pull request.

aldeed:
@ljharb Anything community can do to help get this merged and released? Code appears similar to working code that I had [previously posted](https://github.com/facebook/prop-types/issues/200#issuecomment-409391696) and to the `prop-types-extra` code mentioned, and your earlier comments appear to have been addressed. In general LGTM as far as fixing my #200 issue.

NSLS:
Can we get some attention from the maintainers?

Without this PR we can't get rid of the PropTypes warnings from third-party packages, as they all rely on a component being either string, node or function.

slorber:
Hi,

Not sure it's totally related but I'd like to do this:

` tagName: PropTypes.oneOfType([PropTypes.string, PropTypes.element, React.Fragment]),`

Any idea if this can be done today? it seems fragment is "included" into elementType but can we restrict to fragment?

ljharb:
I don’t think that’s currently possible, no. File a new issue?

Asday:
Missing from the docs.

https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes

ljharb:
@Asday the note at the top of that page tells you more or less to to ignore that page and go to the `prop-types` library, which does have it in the docs.

Asday:
Dangit I'm trying to look at the page now and it's got DNS errors.

I wasn't aware the documentation had moved.  :(

PR #224, opened 2018-10-02
Fix oneOf when used with Symbols

### LINKED ISSUE #10: PropTypes.oneOf relies upon stringification when using Symbols
Original issue: https://github.com/facebook/react/issues/9181

jimf:
- Fixes error due to an attempt to coerce a Symbol to a string
- Improves formatting of the "expected" portion of the generated warning, outputting for example `["Symbol(A)","Symbol(B)"]` rather than `[null,null]`

Fixes #10

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -302,8 +302,13 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
         }
       }
 
-      var valuesString = JSON.stringify(expectedValues);
-      return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` of value `' + propValue + '` ' + ('supplied to `' + componentName + '`, expected one of ' + valuesString + '.'));
+      var valuesString = JSON.stringify(expectedValues, function replacer(key, value) {
+        if (getPropType(value) === 'symbol') {
+          return String(value);
+        }
+        return value;
+      });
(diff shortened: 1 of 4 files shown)


factoryWithTypeCheckers.js
```diff
@@ -302,8 +302,13 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
         }
       }
 
-      var valuesString = JSON.stringify(expectedValues);
-      return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` of value `' + propValue + '` ' + ('supplied to `' + componentName + '`, expected one of ' + valuesString + '.'));
+      var valuesString = JSON.stringify(expectedValues, function replacer(key, value) {
+        if (getPropType(value) === 'symbol') {
```
ljharb: Why not unconditionally use String here?

ljharb: In the template literal they’ll all be converted to a string anyways; it seems simpler to avoid the symbol check.

jimf: For clarity, am I correct that you're suggesting the following?

```js
var valuesString = JSON.stringify(expectedValues, function replacer(key, value) {
  return String(value);
});
```

I just want to point out that this changes the warning output slightly:

```
// Before:
... expected one of ["red","blue"].

// After:
... expected one of "red,blue".
```

Thoughts?

ljharb: oh right, because of objects and arrays. i guess as-is is better, then.

PR #225, opened 2018-10-02
Remove dependency on object-assign

Chalarangelo:
As the code gets processed anyways, the dependency on a polyfill such as `object-assign` is not necessary. I replaced said dependency with an assignment to the native `Object.assign` method, so that the code works exactly like before but with one less dependency.

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -7,7 +7,7 @@
 
 'use strict';
 
-var assign = require('object-assign');
+var assign = Object.assign;
 
 var ReactPropTypesSecret = require('./lib/ReactPropTypesSecret');
 var checkPropTypes = require('./checkPropTypes');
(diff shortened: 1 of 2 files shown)


ljharb:
If the negligible cost of a single dependency is a problem, you can alias it out in your bundler.

jimmywarting:
What if we changed the only line it's used? from

```js
var allKeys = assign({}, props[propName], shapeTypes);
for (var key in allKeys) {
```
to
```js
var allKeys = Object.keys(props[propName]).concat(Object.keys(shapeTypes));
for (var i = 0; i < allKeys.length; i++) {
  var key = allKeys[i]
```

`Object.keys` is supported in more browsers

jimmywarting:
Another tip since you have babel

```js
var allKeys = assign({...props[propName], ...shapeTypes});
for (var key in allKeys) {
```

ljharb:
That would be different; `Object.keys` doesn’t include symbols, and neither does `for..in`.

PR #232, opened 2018-10-23
fixed broken proptypes in prod

### LINKED ISSUE #231: Production build each propType is same instance of shim
We use this `propTypes` for custom elements to make runtime parsing based upon prop-types.

Unfortunately in Production the following is true:

```js
import PropTypes from 'prop-types';

PropTypes.string === PropTypes.bool === PropTypes.number === PropTypes.array === PropTypes.object
// -> true
```

Though this should be false.

Currently the `shim` function is static, It would be better to generate it or use a shim factory.
https://github.com/facebook/prop-types/blob/master/factoryWithThrowingShims.js#L15-L59

AndyOGo:
fixes #231

### DIFF
--- a/factoryWithThrowingShims.js
+++ b/factoryWithThrowingShims.js
@@ -12,39 +12,41 @@ var ReactPropTypesSecret = require('./lib/ReactPropTypesSecret');
 function emptyFunction() {}
 
 module.exports = function() {
-  function shim(props, propName, componentName, location, propFullName, secret) {
-    if (secret === ReactPropTypesSecret) {
-      // It is still safe when called from React.
-      return;
-    }
-    var err = new Error(
-      'Calling PropTypes validators directly is not supported by the `prop-types` package. ' +
-      'Use PropTypes.checkPropTypes() to call them. ' +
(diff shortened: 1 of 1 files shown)


AndyOGo:
I just signed the CLA 🎉

ljharb:
See #106.

LucaMele:
@ljharb , please reconsider your desicion and instead of a quick check that is similar to the 106 try to find out what we would like to achive here. We wont have performance impact with this and we are consistant between DEV environment and PROD environment.

LucaMele:
is that not a win win?

ljharb:
@LucaMele this isn't my decision; i'm not a collaborator here. React team members were pretty clear about it on #106 tho.

LucaMele:
Hello @ljharb all right sorry :). Well hope that the react team memebers can give it a second chance then. I guess we can call it best practice if code does not change logic between PROD and DEV

ljharb:
Although I'd prefer this type of fix to go in, closing per https://github.com/facebook/prop-types/pull/106#issuecomment-398443259.

Issue #181, opened 2018-04-26
Proptypes validation of shape inside an array


oswagner:
We have an array of objects and when try to validade the shape of the object inside that array, it seems to ignore that validation.

```
data: PropTypes.arrayOf(PropTypes.shape({
 someValue: PropTypes.bool
})
```

if we change **PropTypes.bool** in the sample above for **PropTypes.potato**, like below:

```
data: PropTypes.arrayOf(PropTypes.shape({
 someValue: PropTypes.potato
})
```

We don't receive any kind of message about the error.

oswagner:
@ljharb Sorry, but when I gave the example of **PropTypes.potato**. It refers if I try to change to any other type, the same problem occurs, even if I use something like ``PropTypes.number`` 😆

ljharb:
@oswagner see https://jsfiddle.net/69z2wepo/182882/ - I see a propType warning in the console, and when I change "number" to "string", it no longer shows an error.

Are you sure you're using the development build of React? The production build of react doesn't check propTypes.

ljharb:
This might be caused by #220; closing as both a duplicate and answered.

PR #234, opened 2018-10-26
Fix silently failing invalid validators in shape

### LINKED ISSUE #220: No warnings for invalid validators inside shape
This could be related to #181 

In a project we noticed that having typos in prop type definitions that are nested inside a `PropTypes.shape` do not generate warnings in version `15.6.2`:

```js
{
  a: PropTypes.shape({
    b: PropTypes.boolean // <- is undefined
  })
}
```

I've boiled it down to these test cases:

Case 1:
```js
const PropTypes = require('prop-types');

PropTypes.checkPropTypes({ a: undefined }, { a: "hello" });
```

Case 2:
```js
const PropTypes = require('prop-types');

PropTypes.checkPropTypes(
  { a: PropTypes.shape({ b: undefined }) },
  { a: { b: "hello" } }
);
```

Case 1 generates a warning, as expected:

```
Warning: Failed undefined type: React class: undefined type `a` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.
```

Case 2, however, does not output any error or warning 😞 In other words, it seems like if you use an invalid validator inside a `PropTypes.shape`, validation will always pass and no compile time warning or error is visible, which means there might be potential for some sneaky runtime bugs.

I've tested this in many different versions (as low as `0.2.0` and `0.1.0` and it seems to be consistent across all versions.

Is this somehow the intended behaviour? If so, what is a good way to get compile-time errors/warnings for invalid validators used inside `PropTypes.shape`?

asbjornh:
Fixes #220 
Possibly also #181 

This PR adds type checks for validators in `createShapeTypeChecker` and `createStrictShapeTypeChecker` so that they can fail loudly on invalid validators and everyone can be safe and happy! 😊

As outlined in #220 , providing any falsy value instead of a validator will cause validation to be skipped, meaning invalid props won't be caught. Providing a truthy value that's not a function results in a caught but displayed javascript error `Warning: Failed prop type: checker is not a function` which is better than nothing but not really that helpful.

The most likely cause of issues is typos in validator names, like `PropTypes.boolean` (which I've since discovered that I've written several times). As mentioned in #220 this is mitigated by using `eslint-plugin-react/no-typos`, but I do feel that relying on a third party that's not a dependency is more of a band-aid than a solution.

Another cause would be references to internal or external things that might or might not be functions, which is not caught by the eslint plugin.
```js
const validator = false;
Component.propTypes = {
  foo: PropTypes.shape({ bar: validator })
};
```

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -400,6 +400,13 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
     return createChainableTypeChecker(validate);
   }
 
+  function invalidValidatorError(componentName, location, propFullName, key, type) {
+    return new PropTypeError(
+      (componentName || 'React class') + ': ' + location + ' type `' + propFullName + '.' + key + '` is invalid; ' +
+      'it must be a function, usually from the `prop-types` package, but received `' + type + '`.'
+    );
+  }
+
   function createShapeTypeChecker(shapeTypes) {
     function validate(props, propName, componentName, location, propFullName) {
       var propValue = props[propName];
@@ -409,8 +416,8 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
       }
       for (var key in shapeTypes) {
         var checker = shapeTypes[key];
-        if (!checker) {
-          continue;
+        if (typeof checker !== 'function') {
+          return invalidValidatorError(componentName, location, propFullName, key, getPreciseType(checker));
         }
         var error = checker(propValue, key, componentName, location, propFullName + '.' + key, ReactPropTypesSecret);
         if (error) {
@@ -434,6 +441,9 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
       var allKeys = assign({}, props[propName], shapeTypes);
       for (var key in allKeys) {
         var checker = shapeTypes[key];
+        if (has(shapeTypes, key) && typeof checker !== 'function') {
+          return invalidValidatorError(componentName, location, propFullName, key, getPreciseType(checker));
+        }
         if (!checker) {
           return new PropTypeError(
             'Invalid ' + location + ' `' + propFullName + '` key `' + key + '` supplied to `' + componentName + '`.' +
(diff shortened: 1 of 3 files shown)


ljharb:
@gaearon do you think this is a bug fix, or a breaking change because it might introduce additional warnings?

asbjornh:
@ljharb In hindsight I think the tests might be a little thin. I'd be happy to add some more invalid cases in both tests unless that's too late now (I don't fully understand what you did with with those force pushes)

ljharb:
@asbjornh definitely not too late! i rebased the PR branch; after a `git fetch` locally you can `git reset --hard origin/fix-invalid-validators-in-shape`, or `git checkout master && git branch -D fix-invalid-validators-in-shape && git checkout fix-invalid-validators-in-shape`, or you can delete and re-clone your local repo.

asbjornh:
Great! Thanks :)

asbjornh:
@ljharb I've added some more test cases (which was good because I discovered an issue). I think this is ready now, unless you guys want any changes :)

factoryWithTypeCheckers.js
```diff
@@ -434,6 +441,9 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
       var allKeys = assign({}, props[propName], shapeTypes);
       for (var key in allKeys) {
         var checker = shapeTypes[key];
+        if (shapeTypes.hasOwnProperty(key) && typeof checker !== 'function') {
```
ljharb: Please never rely on `hasOwnProperty` being on an object directly; use `has` instead
```suggestion
        if (has(shapeTypes, key) && typeof checker !== 'function') {
```

factoryWithTypeCheckers.js
```diff
@@ -434,6 +441,9 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+          return invalidValidatorError(componentName, location, propFullName, key, typeof checker);
```
ljharb: instead of using `typeof` here, can we use `getPreciseType`?
```suggestion
          return invalidValidatorError(componentName, location, propFullName, key, getPreciseType(checker));
```

factoryWithTypeCheckers.js
asbjornh: 👍

factoryWithTypeCheckers.js
asbjornh: Yep!

asbjornh:
Yay!

PR #245, opened 2018-12-20
Output warnings using console.warn

guilhermearaujo:
Checking the prop types does not change how the program will actually work, and I believe that's why all invalid props are reported with a _Warning_ message.

Yet, these messages are output using `console.error`, and this log level clearly conflicts with the message idea.

This PR replaces all those calls with `console.warn`, bringing consistency to the log levels.

### DIFF
--- a/__tests__/PropTypesDevelopmentReact15.js
+++ b/__tests__/PropTypesDevelopmentReact15.js
@@ -21,20 +21,20 @@ function resetWarningCache() {
 }
 
 function getPropTypeWarningMessage(propTypes, object, componentName) {
-  if (!console.error.calls) {
-    spyOn(console, 'error');
+  if (!console.warn.calls) {
+    spyOn(console, 'warn');
   } else {
-    console.error.calls.reset();
+    console.warn.calls.reset();
   }
   resetWarningCache();
 
   PropTypes.checkPropTypes(propTypes, object, 'prop', 'testComponent');
-  const callCount = console.error.calls.count();
+  const callCount = console.warn.calls.count();
   if (callCount > 1) {
     throw new Error('Too many warnings.');
   }
-  const message = console.error.calls.argsFor(0)[0] || null;
-  console.error.calls.reset();
+  const message = console.warn.calls.argsFor(0)[0] || null;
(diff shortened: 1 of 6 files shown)


ljharb:
It should be an error; if PropTypes are invalid, it means you have a bug.

guilhermearaujo:
Then why does the text say _Warning_? Should it be changed to _Error_?

ljharb:
It’s using console.error because it’s an error; i assume it’s saying “warning” because it doesn’t actually stop you from misusing your components.

guilhermearaujo:
Exactly, it still allow the usage of the components. I agree that an inconsistency in the prop types may end in a _potential_ actual error. But it also may _not_.

JS is not strongly typed and things can work even when the developer is not very thoughtful about their code correctness (but let's not bring this to the discussion).

My argument is that the checking routine does not affect the code execution. The program will still run as it is intended (read: _as the code says to run_). If the validation should report an error, it would make more sense, in my opinion, to throw an actual error or exception, interrupting the execution.

If the intention is to "just let the developer know", a warning does look more appropriate, I believe.

ljharb:
It’s a guaranteed error if explicit PropTypes are invalid - either in the propTypes or in the consumer code. That it still might function correctly by accident doesn’t mean it’s not an error.

console.error is the right place to send this error output.

ljharb:
Closing as a duplicate of #63.

eps1lon:
@ljharb That PR only downgrades the deprecation warning from `console.error` to `console.warn`. This PR downgrades **all**. If anything #63 should be closed over this one.

ljharb:
ah. in that case I'll reopen, since it's not a duplicate.

Gsiete:
In the meantime I'm using:
```
console.error = (function () {
  const { error } = console;

  return function (...args) {
    if (args[0] && args[0].match(/^warn(ing)?:/i)) {
      console.warn(...args);
    } else {
      error.apply(console, args);
    }
  };
}());
```
One particular case where I don't want to see it as an error is: https://github.com/reactstrap/reactstrap/issues/1596

PR #251, opened 2019-01-30
Replace error log with warn

vkhv:
It difficult to find real error in a big project, when you have no way fix components of other teams.


threepointone:
how would making it warn instead fix that problem?

ljharb:
This is a duplicate of #245; also a duplicate of #63, so I'm going to close it.

PR #262, opened 2019-02-27
[Docs] Add instructions for intentional inclusion of validation in production

rgraffbrd:
Add instructions for intentional inclusion of validation in production.


rgraffbrd:
See issue https://github.com/facebook/prop-types/issues/34

PR #268, opened 2019-04-11
Optimize

noscripter:
- <del>[x] extract process.env.NODE_ENV</del>
- [x] more intuitive createChainableTypeChecker condition
- [x] use loose-envify as a devDependency

### DIFF
--- a/factoryWithTypeCheckers.js
+++ b/factoryWithTypeCheckers.js
@@ -206,7 +206,8 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
           }
         }
       }
-      if (props[propName] == null) {
+      // undefined == null and null == null are true
+      if (props[propName] === null || props[propName] === undefined) {
         if (isRequired) {
           if (props[propName] === null) {
             return new PropTypeError('The ' + location + ' `' + propFullName + '` is marked as required ' + ('in `' + componentName + '`, but its value is `null`.'));
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
   },
   "homepage": "https://facebook.github.io/react/",
   "dependencies": {
-    "loose-envify": "^1.4.0",
     "object-assign": "^4.1.1",
     "react-is": "^16.8.1"
   },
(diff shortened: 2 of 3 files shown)


TrySound:
@noscripter NODE_ENV is replaced statically with bundler for browsers. It should not be extracted.

noscripter:
> @noscripter NODE_ENV is replaced statically with bundler for browsers. It should not be extracted.

Thanks for that info, can you provide more info for me to dive into?

TrySound:
This convention is shared across many packages (especially react ones)
https://github.com/rollup/rollup-plugin-replace
https://webpack.js.org/guides/production/#specify-the-mode

noscripter:
> This convention is shared across many packages (especially react ones)
> https://github.com/rollup/rollup-plugin-replace
> https://webpack.js.org/guides/production/#specify-the-mode

Coool, thanks.

factoryWithTypeCheckers.js
```diff
@@ -206,7 +206,8 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
           }
         }
       }
-      if (props[propName] == null) {
+      // undefined == null and null == null are true
+      if (props[propName] === null || props[propName] === undefined) {
```
ljharb: why is this a better change? `== null` is more reliable, since `undefined` can be redefined.

package.json
```diff
@@ -47,6 +46,7 @@
     "eslint": "^5.13.0",
     "in-publish": "^2.0.0",
     "jest": "^19.0.2",
+    "loose-envify": "^1.4.0",
```
ljharb: this must be a runtime dependency, otherwise https://github.com/facebook/prop-types/issues/203#issuecomment-462398632 breaks people.

factoryWithTypeCheckers.js
noscripter: > why is this a better change? `== null` is more reliable, since `undefined` can be redefined.

Thanks @ljharb I found this question on stackoverflow, it looks like a problem related to legacy browsers only, but your review did makes sense.

noscripter: Also according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined), undefined should be un-writable.

package.json
noscripter: > this must be a runtime dependency, otherwise [#203 (comment)](https://github.com/facebook/prop-types/issues/203#issuecomment-462398632) breaks people.

Got it, I thought it was only used for build in the first place.

PR #272, opened 2019-05-09
`checkPropTypes`: add warning for props that not validate in prop-types

Zaynex:
Hi, this `pull request` can help developers to distinguish the props that not validate. 
Here is example:
```
function App() {
  const [node, setNode ] = useState(1);
  return <div onClick={() => setNode(Math.random())}>Hi<Child child={'child'} node={node} /></div>
}

class Child extends PureComponent {
  render() {
    console.log('render');
    return <div>{this.props.child}</div>
  }
}
Child.propTypes = {
  child: PropTypes.number,
}
```
Even if `Child` component is `pureComponent`, it still cause `Child` re-render if unused props change(`node in example`).

### DIFF
--- a/checkPropTypes.js
+++ b/checkPropTypes.js
@@ -13,7 +13,7 @@ if (process.env.NODE_ENV !== 'production') {
   var ReactPropTypesSecret = require('./lib/ReactPropTypesSecret');
   var loggedTypeFailures = {};
   var has = require('./lib/has');
-
+  var diffArr = require('./lib/diffArr');
   printWarning = function(text) {
     var message = 'Warning: ' + text;
     if (typeof console !== 'undefined') {
@@ -86,7 +86,26 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
         }
       }
     }
+
+    var missKeys = checkIfPropsMissKey(values, typeSpecs);
+    if(missKeys){
+      printWarning(
+        'Missing `prop-types`: '+
+        (componentName || 'React class') + ': ' + location + ' type `'+ missKeys +
+      '` is missing validate in `prop-types`. Please add type in `prop-types` or remove '
+      + location + ' `' + missKeys + '` from outer props if not used for performance reason.');
+    }
   }
+}
+
+function checkIfPropsMissKey(componentProps, propTypes) {
+  var componentKeys = Object.keys(componentProps);
+  var propKeys = Object.keys(propTypes);
+  var diffKeys = diffArr(componentKeys, propKeys);
+  if(diffKeys.length) {
+    return diffKeys.join('`, `');
+  }
+  return false;
 }
 
 /**
--- a/lib/diffArr.js
+++ b/lib/diffArr.js
@@ -1,0 +1,3 @@
+module.exports = function diffArr(base, target) {
+  return base.filter(function(item){return target.indexOf(item) < 0});
+}
--- a/__tests__/PropTypesDevelopmentStandalone-test.js
+++ b/__tests__/PropTypesDevelopmentStandalone-test.js
@@ -135,6 +135,42 @@ describe('PropTypesDevelopmentStandalone', () => {
       );
     });
 
+    it('should warn for missing validator', () => {
+      spyOn(console, 'error')
+      const propTypes = { foo: PropTypes.string };
+      const props = { foo: 'foo', bar: 'bar' };
+      PropTypes.checkPropTypes(
+        propTypes,
+        props,
+        'prop',
+        'testComponent',
+        null,
+      );
+      expect(console.error.calls.argsFor(0)[0]).toEqual(
+        'Warning: Missing `prop-types`: testComponent: prop type `bar` is missing validate in `prop-types`.' +
+        ' Please add type in `prop-types` or remove prop' + ' `bar`' +
+        ' from outer props if not used for performance reason.'
+      );
+    });
+
+    it('should warn for missing validators', () => {
+      spyOn(console, 'error')
+      const propTypes = { foo: PropTypes.string };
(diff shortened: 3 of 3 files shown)


lib/diffArr.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function diffArr(base, target) {
+  return base.filter(function(item){return target.indexOf(item) < 0});
+}
```
ljharb:
```suggestion
}

```

lib/diffArr.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function diffArr(base, target) {
+  return base.filter(function(item){return target.indexOf(item) < 0});
```
ljharb:
```suggestion
  return base.filter(function (item) { return target.indexOf(item) < 0; });
```

checkPropTypes.js
```diff
@@ -86,7 +86,26 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
...
+      + location + ' `' + missKeys + '` from outer props if not used for performance reason.');
+    }
+  }
+}
+
+function checkIfPropsMissKey(componentProps, propTypes) {
+  var componentKeys = Object.keys(componentProps);
+  var propKeys = Object.keys(propTypes);
+  var diffKeys = diffArr(componentKeys, propKeys);
+  if(diffKeys.length) {
```
ljharb:
```suggestion
  if (diffKeys.length > 0) {
```

checkPropTypes.js
```diff
@@ -86,7 +86,26 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
...
+    return diffKeys.join('`, `');
   }
+  return false;
```
ljharb:
```suggestion
  return '';
```
should this return an empty string, rather than being "string or false"?

checkPropTypes.js
```diff
@@ -86,7 +86,26 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
         }
       }
     }
+
+    var missKeys = checkIfPropsMissKey(values, typeSpecs);
+    if(missKeys){
```
ljharb:
```suggestion
    if (missKeys) {
```

checkPropTypes.js
```diff
@@ -86,7 +86,26 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
...
+      printWarning(
+        'Missing `prop-types`: '+
+        (componentName || 'React class') + ': ' + location + ' type `'+ missKeys +
+      '` is missing validate in `prop-types`. Please add type in `prop-types` or remove '
```
ljharb:
```suggestion
      '` is missing validation in `.propTypes`. Please add a validator in `.propTypes` or remove '
```

checkPropTypes.js
```diff
@@ -86,7 +86,26 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
...
+      + location + ' `' + missKeys + '` from outer props if not used for performance reason.');
```
ljharb:
```suggestion
      + location + ' `' + missKeys + '` from outer props if not used.');
```

checkPropTypes.js
```diff
@@ -86,7 +86,26 @@ function checkPropTypes(typeSpecs, values, location, componentName, getStack) {
...
+    }
+  }
+}
+
+function checkIfPropsMissKey(componentProps, propTypes) {
+  var componentKeys = Object.keys(componentProps);
+  var propKeys = Object.keys(propTypes);
+  var diffKeys = diffArr(componentKeys, propKeys);
```
ljharb: i think it'd be better to inline this, since it's only used once.

lib/diffArr.js
Zaynex: Yes. But I haven't heard of it before, my pull request and `PropTypes.extra` have the same goal that tell developers should give outer props warning if not validate in  propsTypes. So developers can know how many props actually we use, and remove unnecessary props.

ljharb: right - but i think doing that by default would be quite noisy, and a breaking change, whereas the component author already has the ability via `exact` to achieve the same goal.

Zaynex: well,  I have a question:
If you add `extra` in your project, are you both have `extra` and `no extra` components ?
I think those who notice extra, are sure to want to check all props 

Even we have `React.memo` and `PureComponent`, It will still cause re-render when unsed props change. 

And by the way, what is prop-types aim? It's only check the props you use or check the props from outer?

the `PropTypes.x.isRequired`  warn developer I must have value from parent. why we shouldn't warn developer that props not validated in child?

checkPropTypes.js
Zaynex: I need the function name to comment self .Should I move `checkIfPropsMissKey` in checkPropTypes?

lib/diffArr.js
ljharb: (exact, not extra)

Yes, this is true - generally that's why you'd want to use a linter rule to require all your components use exact props. There are many components - like react-router's `Link` - that would break with this restriction.

Zaynex: My consideration is not enough, thank you for giving me a negative example and your patience. 
Do you think should we add exact feature in prop-types?so we don't have to introduce two libraries. 
```
import PropTypes, { exact } from 'prop-types';
```

ljharb: It’s already there - the separate library just predates PropTypes.exact

Zaynex: Now I think I should close this pr. My careless about the documentation.
thx.

Zaynex:
1. Already implementation in current version.
2. break change the default behavior.

PR #284, opened 2019-08-06
Add key to error messages

### LINKED ISSUE #279: Better prop types warnigs
**Do you want to request a *feature* or report a *bug*?**
Request a feature

**What is the current behavior?**
React prop types warning doesn't reveal info on the component's instance

**What is the expected behavior?**
Print also the component's props

**Which versions of React, and which browser / OS are affected by this issue? Did this work in previous versions of React?**
N/A

The current warnings point the developer only to the source code, and it would be AMAZING if we'll easily see which component instance caused the warnings. Consider a list with hundreds of items, each rendering a component. ATM there's no convenient way to track down the renegade instance.

So instead of the current behavior:

image

Also add:

```
index.js:1375 Warning: Failed prop type: The prop `latitude` is marked as required in `RestaurantListItem`, but its value is `null`.

RestaurantListItem's instance has the following props: {OBJECT_OF_PROPS_THAT_OPENS_ON_CLICK}
```

I currently find myself many times having to temporary do this in different components.

It seems to me like a very easy thing to add, at least for a development build.

Thoughts?

I can take a swing at it with some guidance (I'd love to dip my toes in React's code)

asbjornh:
Fixes #279 

This PR aims to solve the problem of having many failed propTypes checks when rendering a huge list of components. In this case, it can be really hard to determine which warning belongs to which component instance, making debugging tedious and frustrating.

Displaying the `key` in warnings could help to quickly identify which item(s) in a list of components have type warnings:

`Warning: Failed prop type: The prop 'blabla' is marked as required in 'SomeComponent' with key 'abc', but its value is 'null'.`

While this would mitigate the issue, I'm not a 100% sure that this is a good idea (see below), but I'm opening this PR in hopes of at least getting some discussion around it.

This solution should work decently when doing `array.map`, but it comes with at least one side effect that might not be desirable. As far as I know the `key` prop doesn't hold any special significance outside the case of `array.map` and might be used as a part of the regular data flow. In those cases it might be weird that the `key` prop is displayed in warnings. Detecting whether a component was rendered with `array.map` is likely impossible in `prop-types` so I don't see a way of avoiding this.

Also, there might be a need for truncating very long keys to avoid destroying the readability of the message?

### DIFF
--- a/__tests__/PropTypesDevelopmentReact15.js
+++ b/__tests__/PropTypesDevelopmentReact15.js
@@ -251,6 +251,27 @@ describe('PropTypesDevelopmentReact15', () => {
       expectInvalidValidatorWarning(PropTypes.exact({ bar: 'true' }), 'string');
       expectInvalidValidatorWarning(PropTypes.exact({ bar: null }), 'null');
     });
+
+    it('should output the value of the `key` prop (if it has one)', () => {
+      function typeCheckFailWithKey(declaration, value) {
+        const propTypes = { testProp: declaration };
+        const props = { testProp: value, key: 'abc' };
+        const message = getPropTypeWarningMessage(propTypes, props, 'testComponent');
+        expect(message).toContain('`testComponent` with key `abc`');
+      }
+      function Something() {}
+      typeCheckFailWithKey(PropTypes.string, false);
+      typeCheckFailWithKey(PropTypes.arrayOf(PropTypes.string), 2);
+      typeCheckFailWithKey(PropTypes.element, {});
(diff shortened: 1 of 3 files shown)


asbjornh:
Closing this since there hasn't been any activity for a while!

ljharb:
@asbjornh that's not a good reason to close things, but if you're no longer interested in the PR, so be it.

asbjornh:
You're right I should've been more clear. I have sort of lost interest because like I wrote in the description I don't think this is such a good idea anymore :)

PR #289, opened 2019-08-29
mapOf and setOf validators

### LINKED ISSUE #190: .set and .setOf (and .map and .mapOf)
In a project I'm currently working on, I pass [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)s to components.

As a fanatic user of `prop-types` to make components easier to reuse/rearrange, I write `PropTypes.arrayOf(PopTypes.string.isRequired)` quite often. However, there is currently no counterpart for sets. One resorts to defining a prop `PropTypes.object`. This alternative doesn't have quite the positive effect on readability as `.arrayOf`.

I would like to discuss adding `.set` and especially `.setOf`.

I wrote a proof-of-concept. I am also more than happy to provide a pull request including everything from implementation to tests, but figured the wise thing to do is ask for some early feedback.
```javascript
function createSetOfTypeChecker(typeChecker) {
  function validate(props, propName, componentName, location, propFullName) {
    if (typeof typeChecker !== 'function') {
      return new PropTypeError('Property `' + propFullName + '` of component `' + componentName + '` has invalid PropType notation inside setOf.');
    }
    var propValue = props[propName];
    if (!(propValue instanceof Set)) {
      var propType = getPropType(propValue);
      return new PropTypeError('Invalid ' + location + ' `' + propFullName + '` of type ' + ('`' + propType + '` supplied to `' + componentName + '`, expected a set.'));
    }
    var insideValidateResult = null;
    var insideValidateContainer = {};
    propValue.forEach(function checkValue(value) {
      if (null !== insideValidateResult) {
        return;
      }
      insideValidateContainer.value = value;
      var error = typeChecker(insideValidateContainer, 'value', componentName, 'value', 'inside ' + propFullName, ReactPropTypesSecret);
      if (error instanceof Error) {
        insideValidateResult = error;
      }
    });
    return insideValidateResult;
  }
}
```

Errors in this implementation, unfortunately, look like this: ```Invalid value `inside highlightedIds` of type `string` supplied to `NameList`, expected `number`.```

As values inside a set are not a direct property of the set, it is more difficult to name it in the error message. If the fifth value in an array has the wrong type, the error will blab out that it's `array[4]`. In a set, I guess the best thing we could do is say that it's `inside set`.

Has this been discussed before?

conartist6:
Fixes #190

I also intend to add iterableOf to this.


conartist6:
One of the things that I'm sure is worth discussing is the structure of the `mapOf` call, currently `PropTypes.mapOf([keyValidator, valueValidator])`. I thought it was appropriate to use an array, because this will be closely related to `PropTypes.iterableOf`. In fact quite likely I will delegate, making `mapOf([k, v])` internally call `iterableOf([k, v])`. It feels internally consistent to me for these two validators to be related.

factoryWithTypeCheckers.js
```diff
@@ -270,6 +272,74 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+      var keyChecker = typeCheckers[0];
+      var valueChecker = typeCheckers[1];
+      if (typeof Map !== 'function') {
+        return new PropTypeError('Cannot use `PropTypes.mapOf` if the Map data structure is not present');
+      }
+      if (typeof keyChecker !== 'function' || typeof valueChecker !== 'function') {
+        return new PropTypeError('Property `' + propFullName + '` of component `' + componentName + '` has invalid PropType notation inside mapOf.');
+      }
+      var propValue = props[propName];
+      if (!(propValue instanceof Map)) {
```
ljharb: `instanceof` is not actually reliable - it doesn't work cross-realm, and it can be forged (Error can't be brand checked yet, so `instanceof Error` is the best that can be done). To be robust here, we'd have to add a brand check - one complex enough that we'd want to add a dependency (one that i'm in the process of creating anyways, for another use case) that would potentially add to the bundle size.

factoryWithTypeCheckers.js
```diff
@@ -270,6 +272,74 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+  function createSetOfTypeChecker(typeChecker) {
+    function validate(props, propName, componentName, location, propFullName) {
+      if (typeof Set !== 'function') {
+        return new PropTypeError('Cannot use `PropTypes.setOf` if the Set data structure is not present');
+      }
+      if (typeof typeChecker !== 'function') {
+        return new PropTypeError('Property `' + propFullName + '` of component `' + componentName + '` has invalid PropType notation inside setOf.');
+      }
+      var propValue = props[propName];
+      if (!(propValue instanceof Set)) {
```
ljharb: ditto here

factoryWithTypeCheckers.js
conartist6: Did you see [my comment](https://github.com/facebook/prop-types/issues/190#issuecomment-526350505) about that? Also I'm curious to know more about the dependency you're creating. I'd sort of spitballed something of the sort a while back: https://github.com/conartist6/structure-ish. Its aim was never to replace an `instanceof` check though in the sense that it wouldn't detect any cross-realm set, only those that declared themselves in a compatible fashion.

conartist6: Regardless of structure-ish though I though the solutioning I proposed was pretty doable. Would you like me to implement it?

conartist6:
So in order to allow validators to delegate to each other, e.g. `mapOf` delegating to `iterableOf` I'm going to need to do some further work with secret. Currently the design is that the validators themselves [don't have the secret](https://github.com/facebook/prop-types/blob/bef0803ccdfeeeeed0e333e51f7c240d54af23d3/factoryWithTypeCheckers.js#L218), so they can't possibly delegate. I assume the reason they don't have it is because that call site also would feed props into a custom validator, and the intent is to avoid leaking the secret that way.

factoryWithTypeCheckers.js
conartist6: If you're curious about how I would go about really solving the cross-realm problem, my answer is in this issue: https://github.com/conartist6/structure-ish/issues/1. Note that this solution would also support Immutable.js data types. I had put this project on the backburner while awaiting feedback on my proposal to integrate with Immutable, but immutable hasn't been actively maintained for a long time now, so -- so much for that.

ljharb: There's no need for that; the mechanisms are built into the language. I'll post links to the packages once they're published.

conartist6: And which mechanisms are they? You're referring to your technique of calling the size getter on a potential set? If you do that will you also be sure to also check for the existence of an Iterator?

conartist6:
I'm going to add a `tupleOf` validator as well since I need it to implement `mapOf` (partly) as `iterableOf(tupleOf([keyType, valueType]))`.

conartist6:
@ljharb I guess I can't dismiss your review as stale, but I believe I have made the changes that you requested, including your improved method for detecting `Map` and `Set` instances.

Please don't ship this yet though. I have to finish writing the tests. I'm just waiting to do it until I have more of an idea whether you would accept these changes.

conartist6:
OK I take back what I said. All the tests are here. CHANGELOG and README are updated. When you get to looking at this, it is ready to merge from my perspective.

factoryWithTypeCheckers.js
```diff
@@ -269,6 +288,107 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
     }
     return createChainableTypeChecker(validate);
   }
+  
+  function createTupleOfTypeChecker(typeCheckerTuple) {
+    var argsInvalid = !Array.isArray(typeCheckerTuple) || !typeCheckerTuple.every(function (checker) { return typeof checker === 'function'});
```
ljharb:
```suggestion
    var argsInvalid = !Array.isArray(typeCheckerTuple) || typeCheckerTuple.some(function (checker) { return typeof checker !== 'function'});
```

factoryWithTypeCheckers.js
```diff
@@ -269,6 +288,107 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+    return createChainableTypeChecker(validate);
+  }
+
+  function createIterableOfTypeChecker(typeChecker) {
+    function validate(props, propName, componentName, location, propFullName) {
+      if (typeof typeChecker !== 'function') {
+        return new PropTypeError('Property `' + propFullName + '` of component `' + componentName + '` has invalid PropType notation inside iterableOf.');
+      }
+      if (!(Symbol && Symbol.iterator)) {
+        return new PropTypeError('Cannot use `PropTypes.iterableOf` if Symbol.iterator is not present');
```
ljharb: technically this isn't true; you could use `Array.from` exclusively, which can be polyfilled.

factoryWithTypeCheckers.js
```diff
@@ -269,6 +288,107 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+    var validateIterable = createIterableOfTypeChecker(createTupleOfTypeChecker(typeCheckers));
+
+    function validate(props, propName, componentName, location, propFullName) {
+      if (!Array.isArray(typeCheckers)) {
+        return new PropTypeError('Correct syntax is `PropTypes.mapOf([PropTypes.keyType, PropTypes.valueType])`.');
+      }
+      var keyChecker = typeCheckers[0];
+      var valueChecker = typeCheckers[1];
+      if (typeof Map !== 'function') {
+        return new PropTypeError('Cannot use `PropTypes.mapOf` if the Map data structure is not present');
```
ljharb: is this necessary? if `Map` isn't present, `isMap` will fail.

factoryWithTypeCheckers.js
```diff
@@ -269,6 +288,107 @@ module.exports = function(isValidElement, throwOnDirectAccess) {
...
+    }
+    return createChainableTypeChecker(validate);
+  }
+
+  function createSetOfTypeChecker(typeChecker) {
+    var validateIterable = createIterableOfTypeChecker(typeChecker);
+
+    function validate(props, propName, componentName, location, propFullName) {
+      if (typeof Set !== 'function') {
+        return new PropTypeError('Cannot use `PropTypes.setOf` if the Set data structure is not present');
```
ljharb: ditto

lib/forEach.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function forEach(iterable, callback) {
+  var iterator = iterable[Symbol.iterator]();
```
ljharb: i think this utility should use `Array.from`, and fall back to explicit array/string iteration

PR #290, opened 2019-08-30
[Dev Deps] update `eslint`

conartist6:


conartist6:
Upgrades eslint to fix JSX errors in test files.
Updates eslint config to represent how the code is written.
Fixes a few minor issues reported by the linter.

.eslintrc
```diff
@@ -3,13 +3,28 @@
   "parserOptions": {
     "ecmaVersion": 5,
   },
+  "env": {
+    "node": true
+  },
+  "extends": "eslint:recommended",
```
ljharb: adding this is a bit too large a change for this project.

.eslintrc
```diff
@@ -3,13 +3,28 @@
...
+      },
+      "rules": {
+        "no-unused-vars": "off"
+      },
       "parserOptions": {
-        "ecmaVersion": "2019",
+        "ecmaVersion": 2019,
         "ecmaFeatures": {
-          "jsx": true,
+          "jsx": true
```
ljharb: trailing commas are allowed in eslintrc files.

yarn.lock
```diff
@@ -3613,10 +3551,9 @@ wrappy@1:
   version "1.0.2"
   resolved "https://registry.yarnpkg.com/wrappy/-/wrappy-1.0.2.tgz#b5243d8f3ec1aa35f1364605bc0d1036e30ab69f"
 
-write@^0.2.1:
-  version "0.2.1"
-  resolved "https://registry.yarnpkg.com/write/-/write-0.2.1.tgz#5fc03828e264cea3fe91455476f7a3c566cb0757"
-  integrity sha1-X8A4KOJkzqP+kUVUdvejxWbLB1c=
+write@1.0.3:
+  version "1.0.3"
+  resolved "https://plangrid.jfrog.io/plangrid/api/npm/npm/write/-/write-1.0.3.tgz#0800e14523b923a387e415123c865616aae0f5c3"
```
ljharb: using your company's private artifactory registry isn't going to work.

conartist6: Sorry, I had noticed that; thanks for fixing it.

conartist6:
@ljharb I believe that without at least a preset eslint does not execute any rules at all. I checked their docs while making this PR, to be specific. Are you sure your preference is not to execute any rules on the actual sources?

ljharb:
At the moment eslint is primarily there to check for syntax; given that this is a project I'm maintaining but don't have final decisionmaking authority on, it wouldn't be up to me to enable wide swaths of rules at this time.

conartist6:
You couldn't even find some relatively unopinionated set of rules that is possible to enable? It would really do your contributors the favor to be able to point out undefined symbols and assignments to consts and that manner of thing. But my IDE can't even highlight those things because in this project they aren't errors.

conartist6:
I'm also confused as to why you left the various configuration pieces which turn rules off, given that there are no rules.

PR #324, opened 2020-08-24
README: clarify what "optional" means

cben:
Based on all the "should be implicitly optional and not warn without values" tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ class MyComponent extends React.Component {
 
 MyComponent.propTypes = {
   // You can declare that a prop is a specific JS primitive. By default, these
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,7 +65,7 @@ class MyComponent extends React.Component {
 
 MyComponent.propTypes = {
   // You can declare that a prop is a specific JS primitive. By default, these
-  // are all optional.
+  // are all optional - `undefined` and `null` are allowed.
```
ljharb, line 68: in addition, they can be omitted entirely (which is distinct from being present but `undefined`). let's note that too!

PR #335, opened 2021-02-09
Clarify usage of elementType (Closes #334)

### LINKED ISSUE #334: Clarify Usage of elementType
According to the [Usage](https://github.com/facebook/prop-types#usage) section in the README, `elementType` is for a "React element type (ie. MyComponent)".

```js
// A React element type (ie. MyComponent).
optionalElementType: PropTypes.elementType,
```

However, I can also pass a string corresponding to the name of a DOM element (e.g. "span", "div", etc.) without triggering a prop types warning.

Will `elementType` validate the string is a valid DOM element, or can I pass `"bologna"` without triggering a prop types warning?

It's not clear what `elementType` is validating, and what's allowable from the **Usage** section alone.

Can we clarify this?

---

Consider the following example.

**JS Fiddle:** https://jsfiddle.net/h02ka7up/63/

```js
function Test() {
    return (
        <Typography component="span">Text</Typography>
    );
}

Typography.propTypes = {
  component: PropTypes.elementType,
  children: PropTypes.node,
};


function Typography(props) {
  	const Component = props.component || "p";
  	return <Component>{props.children}</Component>;
}


ReactDOM.render(
  <Test />,
  document.getElementById('container')
);
```

gbroques:
See https://github.com/facebook/prop-types/issues/334 for context and motivation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,9 @@ MyComponent.propTypes = {
   // A React element (ie. <MyComponent />).
   optionalElement: PropTypes.element,
 
-  // A React element type (ie. MyComponent).
+  // A React element type (eg. MyComponent).
+  // a function, string, or "element-like" object (eg. React.Fragment, Suspense, etc.)
+  // see https://github.com/facebook/react/blob/master/packages/shared/isValidElementType.js
   optionalElementType: PropTypes.elementType,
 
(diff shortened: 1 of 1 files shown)


ljharb (approved):
Thanks!

PR #342, opened 2021-03-31
Update checkPropTypes.js

kamangir78:
requires a string

### DIFF
--- a/checkPropTypes.js
+++ b/checkPropTypes.js
@@ -17,7 +17,7 @@ if (process.env.NODE_ENV !== 'production') {
   printWarning = function(text) {
     var message = 'Warning: ' + text;
     if (typeof console !== 'undefined') {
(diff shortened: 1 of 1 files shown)


checkPropTypes.js
```diff
@@ -17,7 +17,7 @@ if (process.env.NODE_ENV !== 'production') {
   printWarning = function(text) {
     var message = 'Warning: ' + text;
     if (typeof console !== 'undefined') {
-      console.error(message);
+      console.error(string(message));
```
ljharb, line 20: `string` isn't a function.
```suggestion
      console.error(String(message));
```

ljharb (changes requested):
`console.error` definitely does not require a string - and in this case, it's already guaranteed to be a string since it's created as one two lines up.

PR #364, opened 2021-12-02
Fix branch name

KonstHardy:
Fix branch name (master was renamed to main):

master -> main

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,4 +89,4 @@
 
 ## Before 15.5.0
 
-PropTypes was previously included in React, but is now a separate package. For earlier history of PropTypes [see the React change log.](https://github.com/facebook/react/blob/master/CHANGELOG.md)
+PropTypes was previously included in React, but is now a separate package. For earlier history of PropTypes [see the React change log.](https://github.com/facebook/react/blob/HEAD/CHANGELOG.md)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -89,4 +89,4 @@
 
 ## Before 15.5.0
 
-PropTypes was previously included in React, but is now a separate package. For earlier history of PropTypes [see the React change log.](https://github.com/facebook/react/blob/master/CHANGELOG.md)
+PropTypes was previously included in React, but is now a separate package. For earlier history of PropTypes [see the React change log.](https://github.com/facebook/react/blob/main/CHANGELOG.md)
```
ljharb, line 92:
```suggestion
PropTypes was previously included in React, but is now a separate package. For earlier history of PropTypes [see the React change log.](https://github.com/facebook/react/blob/HEAD/CHANGELOG.md)
```
that way we don't have to hardcode the branch name

README.md
```diff
@@ -84,7 +84,7 @@ MyComponent.propTypes = {
 
   // A React element type (eg. MyComponent).
   // a function, string, or "element-like" object (eg. React.Fragment, Suspense, etc.)
-  // see https://github.com/facebook/react/blob/master/packages/shared/isValidElementType.js
+  // see https://github.com/facebook/react/blob/main/packages/shared/isValidElementType.js
```
ljharb, line 87:
```suggestion
  // see https://github.com/facebook/react/blob/HEAD/packages/shared/isValidElementType.js
```

ljharb (changes requested):

CHANGELOG.md
KonstHardy: Yes, it's better this way. Thanks! :)

ljharb (approved):
thanks!

Issue #111, opened 2017-09-19
Supporting ESM


tbranyen:
Working on native ESM integration with React and finding this to be a non-starter. The error I receive is:

```
🚨   Error: 'default' is not exported by node_modules/prop-types/prop-types.js
https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
lib/component.js (2:7)
1: import process from './util/process';
2: import PropTypes from '../node_modules/prop-types/prop-types';
```

Would it be possible to get an ESM-compatible build of this module? It has zero dependencies so this should be really easy to switch to and then compile to ES5/require during the build step.

ljharb:
I'm not sure why that should be necessary - a native ESM integration that can't pull in CJS modules is cut off from most of the internet :-/

tbranyen:
@ljharb well it's how Chrome 61 works and it's fully connected :smiley: Node has a totally different implementation, but that's not really relevant to implementing a standard ES module to be loaded by a standard ES loader.

ljharb:
How browsers work only applies after a build process runs on the server; if you bundle your entry point into an ESM, then browsers would be able to load the bundle just fine.

frandiox:
@tbranyen I think [rollup-plugin-commonjs](https://github.com/rollup/rollup-plugin-commonjs) was precisely made for this, no need to change the whole internet.

tbranyen:
@frandiox, that is condescending and not relevant to this issue. If Facebook doesn't want to use a standard format, that's fine. I'll just load it as UMD and grab the global. Eventually this module will change though. Just a matter of time.

ljharb:
Are you suggesting they use an .mjs file? That’d be the standard format to follow.

tbranyen:
> With the CommonJS require() and module.exports, it is easy to import a function that doesn’t really exist, and not realize that until you call it. However, tools like Rollup that natively support import and export syntax fail the build if you mistype a named import. After releasing React 16, we have converted the entire React source code to the ES Modules syntax.

https://reactjs.org/blog/2017/12/15/improving-the-repository-infrastructure.html#migrating-to-es-modules

Looks like React went the route I suggested, which is great, I'm hoping now a contributor can comment on whether or not a PR would be accepted to land ESM.

mishaor:
still nothing :(

ljharb:
[...]

PR #386, opened 2022-10-16
rushed esm outline

brandonros:
part of https://github.com/facebook/prop-types/issues/111

allows for 

```html
<!doctype>
<html>
  <head>
  </head>
  <body>
      <script type="importmap">
      {
        "imports": {
          "d3": "https://unpkg.com/d3@7.6.1/src/index.js?module",
          "c3": "https://unpkg.com/c3@0.7.20/c3.esm.js",
          "preact": "https://unpkg.com/htm/preact/index.mjs?module",
          "preact-compat": "https://unpkg.com/preact-compat@3.19.0/dist/preact-compat.es.js?module",
          "prop-types": "https://cdn.jsdelivr.net/gh/brandonros/prop-types@esm/esm/index.js"
        }
      }
    </script>
    <script type="module">
      import c3 from 'c3'
      import { html, Component, render } from 'preact'
      import PropTypes from 'prop-types'
      import { findDOMNode } from 'preact-compat'
      console.log({
        c3
      })
    </script>
  </body>
</html>
```

i'm not saying we should actually merge this, just hopes it helps somebody else out there


ljharb:
This definitely isn't a good approach to provide native ESM modules.

What I'm wondering is why you're not bundling anyways - there's tons of performance benefits to using bundles that simply aren't available with native ESM.

kungfooman:
Wow, I'm sick of projects not shipping proper ES6 files.

This PR is a giant time saver, thank you so much! I was about to do the same thing for my project :sweat: :sweat_drops: :face_with_head_bandage: 

> What I'm wondering is why you're not bundling anyways - there's tons of performance benefits to using bundles that simply aren't available with native ESM.

It's mostly about developing and debugging. Once that is done, you can create sweat, fast, mangled, compressed, tersed, tree-shaked and whatnot bundle... but don't eat that thing during development!

**And especially don't provide that as the only option while feeling righteous about it!**

I suggest to make ES6 a top-priority for a while - learn from people who know how to ship NPM packages, like the Monaco team.

PR #387, opened 2022-10-18
Fix bad variable name

Rovel:
Hi, by accident I found this ENV var I thought that the name of the variable could have a better name considering its open source now
I would love to have the `hacktoberfest` tag on this PR if it makes sense merging

### DIFF
--- a/lib/ReactPropTypesSecret.js
+++ b/lib/ReactPropTypesSecret.js
@@ -7,6 +7,6 @@
 
 'use strict';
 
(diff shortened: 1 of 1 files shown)


ljharb:
No thanks, there's no value in changing this.

Rovel:
It's sad that you don't see value in changing a variable that contains a threat in it to become something less threatening, especially since the threat doesn't make sense for 99% of the people that run this code in 2022.

ljharb:
It's intended to be maximally threatening.

PR #393, opened 2023-02-11
Readme to Function Components

Levin779:
Since its recommend to define components as functions instead of classes this should be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,10 +57,8 @@ documents the different validators provided:
 import React from 'react';
 import PropTypes from 'prop-types';
 
-class MyComponent extends React.Component {
-  render() {
-    // ... do things with the props
-  }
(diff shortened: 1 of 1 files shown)


ljharb:
Where is that unconditionally recommended?

ErrorBoundary components still must be class components, and there's a number of patterns that remain clearer/cleaner as class components than SFCs with hooks.

Levin779:
Its recommended in the official react docs: “We recommend to define components as functions instead of classes.“ 
[https://beta.reactjs.org/reference/react/Component](https://beta.reactjs.org/reference/react/Component)

ljharb:
Once those docs are out of beta, this seems like a reasonable change.

wojtekmaj:
Well, docs are now out of beta. What do we do?

PR #396, opened 2023-04-19
[new] add `PropTypes.literal`

wojtekmaj:
Currently, to check if value is exactly something, we need to do it in a way that feels a bit hacky:

```
PropTypes.oneOf([value])
```

With this proposal, one could use:

```
PropTypes.literal(value)
```

to achieve the same thing.


ljharb:
I'm not sure the benefit here is worth the complexity of adding an entirely new propType.

